PR #6, opened 2024-03-19
CI fix for Github Actions

mche3kek:
Upload of missing jar file from template project

PR #10, opened 2024-03-21
Authentication with Google 

FacerOfGod:
- Access the google-service.com throught github secrets.
- Authentication interface with functional Google Log-in button.
- Build test for the coverage.


fdejean:
There is an issue with the token "default_web_client_id" that need to be generated by the google-json. 
In the bootcamp, we could just create a field in string.xml with the key, however this file would be on the github and exposing the secret id.

FacerOfGod:
In the commit "[bug fix](https://github.com/feedme-swent-epfl/feedme-android/pull/10/commits/fdc5db3d0f2569d721182bdb29bb6a21f30e4a93)" I removed the line Firebase.auth().signOut() for now. This line was supposed to logout the current user. This will be done with a future classe FirebaseDatabase

app/build.gradle.kts
mche3kek, line 1: Seems like there are duplicate imports in testImplementation in the app/build.gradle.kts, roboeletric and other testing libraries are imported using the libs.versions.toml file and the lib name + version : 

e.g:     testImplementation(libs.robolectric) **and**   testImplementation("org.robolectric:robolectric:4.11.1")

Could be a good idea to include all libs in the corresponding file

app/src/main/java/com/android/feedme/MainActivity.kt
mche3kek, line 1: Code coverage doesn't test the Greeting and GreetingPreview functions, which are also deprecated as we have a login screen now --> shouldn't we delete them ?

mche3kek (changes requested):
Code coverage is only 10.5%, is it due to the fact that we don't test the UI elements ? Otherwise I think it would be pertinent to improve the build.gradle.kts file and adapt the MainActivity.kt file as well per their comments.

app/src/main/java/com/android/feedme/MainActivity.kt
FacerOfGod: Yes, we could do it now. In either case, it will be deleted when we will properly integrate navigation in the MainActivity

ohamz (approved):
Tested on local device - Google Login works successfully.
Code implementation approved.

ohamz:
@mche3kek implement your requested changes.

fdejean:
The issue seems to be that SonarCloud is not running the androidTest and only the unitTest. 
Is this normal or a bug ? @mche3kek @FacerOfGod

fdejean:
@mche3kek Sonarcloud is not working correctly with the CI, the `Upload report to SonarCloud` at line 55 contain an error.
This error prevent the upload of the accurate coverage, hence the Code Analysis fail.
I am not really sure how to fix it, I tried to upload Jacoco version but it did not work...
#6

fdejean:
I found an article having exactly the same bug. [Here](https://medium.com/@theilacker/fixing-sonarqube-line-out-of-range-issue-when-using-jetpack-compose-5ba4c1f361f1)
_JaCoCo is still unable to properly cover composable functions, reporting multiple invisible lines and branches as missing in the final report, since Compose generates a lot of bytecode behind the scenes and JaCoCo doesn’t offer any filter to hide those in the report_

fdejean:
The fix of the article worked. 
We are missing 5% coverage.
@FacerOfGod can you implement another test ?

app/build.gradle.kts
fdejean: Yes, I cleaned up the gradle to use the  libs.versions.toml file !

ohamz (approved):
Authentication works from my side,  additional tests may be required later.
Approved, PR can be merged.

PR #22, opened 2024-03-31
Navigation Actions UI

ohamz:
Implemented the UI for the NavigationActions file.
Please take a look at it, the Icons may need to be changed later.


FacerOfGod (commented):
NavAction has been added just like the bootcamp. It has been correctly implemented. Changes maybe necessary in the future in case we change the TopLevelDestination.

However if we like to navigate to a non TopLevelDestination, for example : ViewRecipe how do we do that?

fdejean:
Simply, we could add more `Route` and define them with the `composable` in the `MainActivity`.
But this approach is problematic when those views are nested and have no purpose outside of a specific view.
If this is a case, we need to implement a new Nav, the same way we did, inside of the page.
@FacerOfGod

fdejean (changes requested):
The code is correct, but it is missing the tests and the implementation
`navigateTo` must be tested
`MainActivity` has not been updated to implement it
`BottomNavigationMenu`, the same way as  in the bootcamp, would allow to quickly do it

ohamz:
@fdejean @FacerOfGod I suggest we keep the TopLevelDestinations for the screens accessible from the BottomNavigationBar.
Take a look at the link below, I think this is how our MainActivity should look like, a mix of composable and navigation components in order to represent our full nested graph. What do you think ?
https://developer.android.com/guide/navigation/design/nested-graphs#compose

ohamz:
We need to test the navigation system and the BottomNavigationMenu, how do we do it?

FacerOfGod:
> We need to test the navigation system and the BottomNavigationMenu, how do we do it?
BottomNavigationMenu is a composable, so add some TestTags to ensure elements appear.
For NavActions, create an instance and attempt to navigate through different destinations.

ohamz:
> > We need to test the navigation system and the BottomNavigationMenu, how do we do it?
> 
> BottomNavigationMenu is a composable, so add some TestTags to ensure elements appear. For NavActions, create an instance and attempt to navigate through different destinations.

Is it better to wait so that we implement the UI for the 5 main screens ? So that we don't write some tests to later remodify them. (for Navigation tests)

FacerOfGod:
> > > We need to test the navigation system and the BottomNavigationMenu, how do we do it?
> > 
> > 
> > BottomNavigationMenu is a composable, so add some TestTags to ensure elements appear. For NavActions, create an instance and attempt to navigate through different destinations.
> 
> Is it better to wait so that we implement the UI for the 5 main screens ? So that we don't write some tests to later remodify them. (for Navigation tests)

Mmmh you can do it now without UI interaction, but that would be a UnitTest :/ I guess you can wait for the UI and test it like we did in the bootcamp.

FacerOfGod:
Have you had a chance to run any tests?

ohamz:
@fdejean this PR is ready to be merged.
We can also delete the branch and recreate a new one (from the evolved main) when testing will  be required.

fdejean:
> Why did CI fail ? I just resolved conflicts with main so that it's easy to merge.

A rerun fixed the error, @ohamz

fdejean (approved):
Good, but the implementation on `MainActivity`  remains to be done.

FacerOfGod:
@ohamz You can check TopBarNavigationTest and do something familiar to test BottomNavigationMenu

ohamz:
> @ohamz You can check TopBarNavigationTest and do something familiar to test BottomNavigationMenu

Will check out your PR!

PR #23, opened 2024-04-01
Implementing the Database and Data Model

fdejean:
The goal is to get usable Recipe and Ingredients objects for the UI.
More detail about the data model can be found on the wiki / Notion.


fdejean:
The test fail since the Authentication is not currently done, hence it cannot access the database.

fdejean:
I will only implement the UI such that the object are already define for the UI implementation

app/build.gradle.kts
mche3kek, line 1: Instead of having id for the plugins, adding them to the libs file would be more suitable (especially for version management)

build.gradle.kts
mche3kek, line 1: Same as for the other build.gradle.kts file, btw I can do it if you prefer

app/src/main/java/com/android/feedme/model/data/Ingredient.kt
mche3kek, line 1: Maybe add a comment to explain what each field means, also might be interesting to use an enum for the ingredients type as they should be well-defined

mche3kek, line 1: Also for the id, is it only an internal component hidden from the user? If that's the case, wouldn't it be easier to have it as an integer for ease of comparison and such?

app/src/main/java/com/android/feedme/model/data/Recipe.kt
mche3kek, line 1: Same remark for the recipeId regarding the ease of comparison

mche3kek, line 1: I think that the detailedDescription component is not that meaningful (or could be replaced with a Step object).

It could be interesting to have a numberOfSteps / numberOfIngredients component for easier retrieval of these values for the short description of each recipe (e.h : This recipe contains x steps) or for the recipe suggestions module (e.g : you have x ingredients and you still need y)

Maybe change the val shortDescription to title in step ?

For the metadata, I think that measure should be in an enum as it can only be one of few select values, and quantity should be expressed as a number. Also, we can have the measure enum be specific to the type of the ingredient (e.g: if spice --> pinch, tbsp, etc)

mche3kek (changes requested):
I think that the model should be improved upon to avoid having inexact or unusable values for certain fields, I stay at your disposal to discuss the model implementation if need be

app/src/main/java/com/android/feedme/model/data/Recipe.kt
mche3kek: Typo in difficulty

Time in a string will be tricky to parse, maybe a numeric value in minutes

Same remark for userid as the other ids

mche3kek (changes requested):

app/src/main/java/com/android/feedme/model/data/Ingredient.kt
fdejean: Ingredient will not be well defined, as it will be hard to get a total list of them. I will add more comment too!

fdejean: Yes, it is internal, but it is best practice to use a String for Firebase

app/src/main/java/com/android/feedme/model/data/Recipe.kt
fdejean: > I think that the detailedDescription component is not that meaningful (or could be replaced with a Step object).
The short description is for the thumbnail, but I agree that detailedDescription is not used in the Figma. Hence I will remove then
> It could be interesting to have a numberOfSteps / numberOfIngredients component for easier retrieval of these values for the short description of each recipe
I agree for numberOfSteps, but for the numberOfIngredients, this can simply be get by counting the element in the list. 
> For the metadata, I think that measure should be in an enum as it can only be one of few select values, and quantity should be expressed as a number.
Yes, good idea

build.gradle.kts
fdejean: Yes, I will prefer that you do it @mche3kek. As I do not why some are specified and other not

mche3kek:
Everything looks fine to me, model might need tweaking after we begin scrapping recipes

PR #26, opened 2024-04-03
Camera Screen UI

ohamz:
Implemented UI for Camera Screen
- Being able to open the Camera
- Being able to snap a picture
- Being able to select a picture


ohamz:
- Works on simulator, needs testing locally on Android device (@FacerOfGod or @fdejean please if you can :) just run the current version of the branch).
- Need to implement tests.
- Need to fix gradle.kts organization.

ohamz:
Need to see why did CI fail...

fdejean:
> Need to see why did CI fail...

The CI fail because de test does not pass. Do they pass in local ? @ohamz

ohamz:
> Do they pass in local ?

I actually forgot to test it locally 😆
I'll try to fix it.

ohamz:
@fdejean I fixed Camera tests, I needed to grant permission for camera in tests.
One problem is how do we test both LoginScreen and CameraScreen if we call just one of them in MainActivity, this should be fixed when navigation will be implemented.
Also do you know why the SonarCloud check doesnt appear ?

ohamz:
I don't know why some tests aren't passing on CI ...
All is working perfectly locally.

ohamz:
FINALLY it works :)
Should we review, merge and close this PR ?

ohamz:
> * Also, what do you think about holding off on merging until Navigation is online?

I think it's better to merge it so that we don't have dangling branches that are waiting to be merged for no specific reason, they would get accumulated and it would become harder to merge later. Especially for the little changes we do (for example in the gradle).

I will fix the documentation

mche3kek (approved):
As Sylvain said, the Camera.kt would benefit from further documentation, otherwise great work !

ohamz:
> The current version only display the login screen, even upon login-in. This is a bug @ohamz

I don't think it's a bug. In the MainActivity I'm setting the initial state to LOGIN, it then only changed within the tests not when running the app module. If you want to test locally on the emulator you would have to change the initial state to CAMERA.

ohamz:
@FacerOfGod could you recheck the new modifications before merging.

More coverage would be added later when adding tests.

image

ohamz:
@FacerOfGod It's ready to be merged, should be direct without conflicts.

FacerOfGod (approved):

PR #33, opened 2024-04-07
APK building in Github Actions

mche3kek:
Set up a CD to build a new APK release each Sprint


mche3kek:
APK building and draft release creation work as expected, I kept the current branch in the apk.yml file for testing purposes during the review. 

The APK that will be generated has name format of "feedme-android-v1.x.x.apk", with a proposed scheme of : we add 1 at the right-most digit at each feature developed, and we add a 1 at the middle digit at the end of each sprint. 

There is currently a draft release (left untouched after automatic generation) for reference.

fdejean (approved):

PR #39, opened 2024-04-08
Added Profile and Comment Model!

fdejean:
Created Profile and Comment for the UI as described by Figma and @xrsula


xrsula (approved):
Everything looks good ! The Profile attributes will eventually have to be tweaked, but it's good so far

PR #40, opened 2024-04-08
Database Implementation of the Data Model

fdejean:
The goal is to create a basic Database class to easily be able to get and set every objects 

- [x] Profile, 
- [x] Recipe,
- [x] Comments,
- [x] Ingredients


fdejean:
TODO

- [ ] Failure Check
- [ ] Real Test / Upload to Firebase
- [ ] Simple UI

fdejean:
Real Test / Upload to Firebase are done, they do work!!
 I created some dummy Screen to test it. I will not commit them but will share the basic code on Notion with some explanation for a quick implementation.
 Next TODO is to complete the test to have 80% coverage with failure handling too

fdejean:
Why the CI not running ? Is it because of the conflict ?

ohamz:
Florian the BOSS.

I just have a few remarks : 
- In Recipe.kt, in data class Recipe, wouldn't be more pertinent to create a Difficulty enum class for the difficulty with [EASY, MEDIUM, HARD}
- In profile.kt, did you mean to name this file Profile.kt ?
- I guess it was intended for the tests to be in the 'test' folder and not 'androidTest' since they're unit tests.

Other than that, I fully approve this PR, huge effort from your part thank you! I will keep it 'Review Required' so that someone else reviews it (will approve it tomorrow if no one does).
@fdejean

app/src/main/java/com/android/feedme/model/data/Comment.kt
JeannedeM, line 1: Maybe we should give the possibility to people to upload an image of what they have just cooked in their comment. I think it can give good informations to other users about the recipes. Hence you could maybe add a "val imageUrl : String" in the data class Comment.

JeannedeM (approved):
Overall it seems perfect to me ! You did a really great job ! I have nothing to add, excetpt maybe the comment i put about the data class Comment, where i think it could be great to add a "val imageUrl : String" to give users the possibility to upload images in their comments.

app/src/main/java/com/android/feedme/model/data/Comment.kt
JeannedeM: Oh yes i'm sorry ! Indeed github review didn't show me the complete class.... So forget my comment everything is good then !

JeannedeM:
About Omar suggestion to add a Difficulty enum class for the difficulty with [EASY, MEDIUM, HARD}, i think we should wait to see wich recipes data we are using in our database. The recipes difficulty could be rated in differents ways depending on the data, and we could have some trouble to adapt it to our format. So i think we sould wait a bit longuer, and eventually change this detail later :)

fdejean:
>     * In profile.kt, did you mean to name this file Profile.kt ?

No, nice catch! Thanks
>     * I guess it was intended for the tests to be in the 'test' folder and not 'androidTest' since they're unit tests.

Yes, they are unit test and does not require an android emulator to run

PR #41, opened 2024-04-09
36 create UI for recipe small thumbnails

JeannedeM:
Requirements : 

- Display recipes like in the figma for the profile page
- The exact aesthetic (ie. the definitive graphic design, color palet, etc) is not completly fixed in figma, hence some work on the visual aspect of it will be done later in an other issue.
- Each Recipes is rrepresented by Image + Rating + Rating icon + Time + Time icon + Save Button
- Recipes are displayed two by two in a column fashion

image


0marwa (approved):
To me it's looking good :)

PR #46, opened 2024-04-11
Linking the profile data to profile screen

xrsula:
This pull request deals with the linking of the profile data and the profile screen. It takes from the database all information stored about the user that is signed in, and displays it in the profile page in a correct way.

In order to do so, several additional implementation had to be done: 
- the implementation of the data class Profile.kt helps give a skeleton of the profile of each user, with all of its properties,
- the implementation of ProfileViewModel and AuthViewModel permits the extraction of information stored in the database, which is stored following the data class initialised above,
- the modification of ProfileScreen.kt accommodates the ProfileViewModel and permits the extraction of the user's info from the database,
- the modification of LoginScreen.kt accommodates the AuthViewModel and registers the login information of the user, and creates their basic profile
- the modification of MainActivity.kt accommodates the creation of the profile page, and the navigation needed to access it


FacerOfGod (changes requested):
Nice Job! I like the aesthetic of the profile box. However, there are some important modifications that need attention:

- Ensure documentation, like in `LoginScreen`, isn't deleted. _(Note: I actually don't see why this class has been modified ?)_
- Add documentation for `AuthViewModel` and `ProfileViewModel` for clarity.
- Include missing parameter documentation in `ProfileScreen` for better guidance.
- Update default values in `data class Profile` to clearly signify that they are defaults values.
- Provide clearer descriptions for TODOs to guide future work effectively :)

Great job on the recent updates!

fdejean:
@FacerOfGod found a bug on his Emulator on the login. Hence we will roll back the login previously used. 
This would prevent the new User to not be linked with database. Hence the profile display will be the default has it is not in the database.

FacerOfGod (approved):
Nice changes and nice work! In the profile page I can see my name :D

PR #48, opened 2024-04-11
Fix APK building 

mche3kek:
APK generated was the release one, we couldn't install it on phones (my guess would be because of incorrect app signature), we switched to debug APK which is uploaded in the Releases section of our Github


mche3kek:
PR ready to be merged @xrsula

xrsula (approved):
Good job noticing that you had to add the APK to the workflow artifacts ! I tested and it seems to work

PR #51, opened 2024-04-11
TopBarNavigation Bug Fix 

FacerOfGod:
Previously, the TopBarNavigation component lacked a fixed height, resulting in it being inadvertently compressed by other elements added beneath it. Additionally, the title within the navigation bar suffered from readability issues due to its lack of emphasis.
- TopBarNavigation now has a fixed height
- Title is now nold for better readability


ohamz (approved):
Changes approved, nice catch.

PR #53, opened 2024-04-12
44 implement UI for comments small thumbnails on profile page

JeannedeM:
Implements UI to display the comment on the profile page.
Acceptance Criteria : 

- Comments displayed in a Column fashion
- Each Comments card has : image + author name + title + description

image

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -141,6 +141,8 @@ dependencies {
     globalTestImplementation(libs.androidx.espresso.core)
     // Android Navigation
     implementation(libs.androidx.navigation.runtime.ktx)
+    // Import image from url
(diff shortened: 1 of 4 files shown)


FacerOfGod (changes requested):
Nice job! There are just a few changes that are need :
- For good practice add an argument Modifier if your function is a composable.
- In your documentation : 
https://github.com/feedme-swent-epfl/feedme-android/blob/b251358b9fe169f30275b2b6acaf2ebe51e5746e/app/src/main/java/com/android/feedme/ui/navigation/SmallCommentsDisplay.kt#L55
When referencing specifique classes put them in brackets as such [Comment], It makes that documentation easier to use

For later down the line:
- UI should look more like the figma

FacerOfGod (approved):

PR #55, opened 2024-04-12
Implementing the Search Bar + updating the Landing Page functions

0marwa:
I finished implementing the UI for the Search Bar, as well as the tests. 
The functionalities such as actually searching and filtering are still undefined though.

### DIFF
--- a/app/src/main/java/com/android/feedme/ui/LandingScreen.kt
+++ b/app/src/main/java/com/android/feedme/ui/LandingScreen.kt
@@ -81,15 +81,12 @@ fun LandingPage(navigationActions: NavigationActions) {
               "https://external-content.duckduckgo.com/iu/?u=https%3A%2F%2Fwww.mamablip.com%2Fstorage%2FLasagna%2520with%2520Meat%2520and%2520Tomato%2520Sauce_3481612355355.jpg&f=1&nofb=1&ipt=8e887ba99ce20a85fb867dabbe0206c1146ebf2f13548b5653a2778e3ea18c54&ipo=images"),
       )
   Scaffold(
-      modifier = Modifier
-          .fillMaxSize()
-          .testTag("LandingScreen"),
+      modifier = Modifier.fillMaxSize().testTag("LandingScreen"),
       topBar = { TopBarNavigation(title = "FeedMe") },
       bottomBar = {
         BottomNavigationMenu(Route.HOME, navigationActions::navigateTo, TOP_LEVEL_DESTINATIONS)
       },
-      content = { padding -> LandingScreenComplete(padding, testRecipes) }
-  )
+      content = { padding -> LandingScreenComplete(padding, testRecipes) })
 }
 
 /**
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/android/feedme/ui/LandingScreen.kt
```diff
@@ -67,7 +75,7 @@ fun LandingPage(navigationActions: NavigationActions) {
               tags = listOf("Meat"),
               time = 1.15,
               rating = 4.5,
-              userid = "PasDavid",
+              userid = "EtienneBoisson",
               difficulty = "Intermediate",
               "https://external-content.duckduckgo.com/iu/?u=https%3A%2F%2Fwww.mamablip.com%2Fstorage%2FLasagna%2520with%2520Meat%2520and%2520Tomato%2520Sauce_3481612355355.jpg&f=1&nofb=1&ipt=8e887ba99ce20a85fb867dabbe0206c1146ebf2f13548b5653a2778e3ea18c54&ipo=images"),
```
SekoiaTree, line 80:
```suggestion
              "https://example.com"),
```
This is a very long URL that can just be an example URL.

(Also, why are testing things mixed in with real code?)

app/src/main/java/com/android/feedme/ui/LandingScreen.kt
```diff
@@ -184,3 +191,66 @@ fun RecipeList(padding: PaddingValues, recipes: List<Recipe>) {
...
+ * Function that displays our search bar as well as the filters icon
+ *
+ * @param searchText : the default text displayed in the search bar
+ * @param onSearchTextChanged : if we want to apply any changes to the text we type in the search
+ *   bar
+ * @param onFilterClicked : the function that will allow us to go through filters by clicking on the
+ *   icon
+ */
+@Composable
+fun SearchBar(
```
SekoiaTree, line 205: Shouldn't this be a https://developer.android.com/reference/com/google/android/material/search/SearchBar instead?

app/src/main/java/com/android/feedme/ui/LandingScreen.kt
```diff
@@ -184,3 +191,66 @@ fun RecipeList(padding: PaddingValues, recipes: List<Recipe>) {
...
+
+/**
+ * Function responsible for the general landing page Contains namely the Search Bar and the Recipes
+ * List
+ *
+ * @param padding : the padding because this will be our main component
+ * @param recipes : the argument to our RecipeList() function
+ */
+@Composable
+fun LandingScreenComplete(padding: PaddingValues, recipes: List<Recipe>) {
```
SekoiaTree, line 247: Nitpick: instead of PaddingValues, you should probably pass a modifier that you then add a testTag and background to?

SekoiaTree (changes requested):
The tests are good!

JeannedeM (changes requested):
Nice job overall, thnaks for all the work ! Here are some changes that should be done : 

- give more details in the documentation of your functions, don't hesitate to add [] around object taken into arguments to link them.
- In your first function please don't keep the fake recipe val. It shoul only be in the test class for test pruposes.
- In your RecipeList function can you give more information of what your first TODO should do.
- It's not really good practice to use a custom fuction SearchBar, one already exist in jetcpack compose.
- Also i'm not sure to understand why you have paddings as arguments in several function.
- and finally as discussed together, you should adapt the UI of thelanding page according to the figma Main page

I will review more carefully your cod after those changes

0marwa:
A few remarks for the reviewers : 
- The search is only partly functional, as well as the filtering icon. These issues are TBD next Sprint
- The UI of the Landing Page still needs some work, which I will be doing on the next Spring
- the `val testRecipes` is temporary, awaiting the `ViewModel` of `Recipe` class (TBD next Sprint too)

Please keep those in mind when reviewing :)

ohamz (approved):
Search Bar implemented as practiced in the bootcamp, and overall tests for LandingPage pass with enough coverage.
Approved - Can be merged.

JeannedeM (approved):

PR #72, opened 2024-04-12
Self signing the APK

mche3kek:


mche3kek:
We close this PR as it's not relevant as of right now

PR #73, opened 2024-04-13
Follower/Following UI

fdejean:
Code the Figma implementation with static content inside of the app


fdejean:
Here is what the UI look like
image
Feel free to provide feedback or modification in any way

xrsula (approved):
Everything looks good, the code is coherent with the functionalities added. The code is also well documented.
The only thing to fix would be the test in LoginTest that makes the CI fail, but it can be fixed at a later date.
Well done !

0marwa (approved):
All good for me! (late review sorry it didn't get posted on time)

PR #74, opened 2024-04-15
66 code logic to edit user profile info

fdejean:
This PR aim at making the Screen and the Logic to modify the User profile

### DIFF
--- a/app/src/main/java/com/android/feedme/MainActivity.kt
+++ b/app/src/main/java/com/android/feedme/MainActivity.kt
@@ -46,7 +46,7 @@ class MainActivity : ComponentActivity() {
               // Navigation host for the app
               val navController = rememberNavController()
               val navigationActions = NavigationActions(navController)
-              val profileViewModel:ProfileViewModel = viewModel()
+              val profileViewModel: ProfileViewModel = viewModel<ProfileViewModel>()
 
               // Set up the navigation graph
               NavHost(navController = navController, startDestination = Route.AUTHENTICATION) {
--- a/app/src/main/java/com/android/feedme/MainActivity.kt
+++ b/app/src/main/java/com/android/feedme/MainActivity.kt
@@ -40,7 +40,9 @@ class MainActivity : ComponentActivity() {
       feedmeAppTheme {
         // A surface container using the 'background' color from the theme
         Surface(
-          modifier = Modifier.fillMaxSize().semantics { testTag = C.Tag.main_screen_container },
+          modifier = Modifier
+            .fillMaxSize()
+            .semantics { testTag = C.Tag.main_screen_container },
           color = MaterialTheme.colorScheme.background) {
           // Navigation host for the app
           val navController = rememberNavController()
@@ -57,10 +59,12 @@ class MainActivity : ComponentActivity() {
             composable(Route.CREATE) { CreateScreen(navigationActions) }
             composable(Route.PROFILE) { ProfileScreen(navigationActions = navigationActions,profileViewModel) }
             composable(Route.SETTINGS) {
-              EditProfileScreen(navigationActions,profileViewModel)
-            }
+              NotImplementedScreen(navigationActions, Route.SETTINGS)             }
             composable(Route.CAMERA) { CameraScreen(navigationActions) }
 
+            composable(Route.EDITPROFILE){
+                EditProfileScreen(navigationActions,profileViewModel)
+            }
 
           }
         }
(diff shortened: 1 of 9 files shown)


fdejean:
Here is a first version of UI with functional logic and safety for

### Name :
- cannot be empty
- must be at least 3 char
- must at most 15 char

### Username
- cannot be empty
- must be at least 3 char
- must be at most 15 char
- must not contain empty space

### Bio
- must be at most 100 char


image
image

fdejean:
Next to do 

- [ ] Add automatic scroll when typing for the Bio, currently we cannot see well what we write
- [ ] Test
- [ ] Documentation

fdejean:
There is an issue with the CI, as the test perform a click action to test the profile edit. 
This click action pass in local and lead to 95% coverage of EditProfileScreen.kt

fdejean:
Tried to manually control the clock of the test to add more time for the action to terminate the test.
The test `testSaveButton` still pass but the coverage does not appear on sonarcloud.

fdejean:
The Bug is not related to the time the action take, it is simply because the Save button is not displayed when the CI runs the test. This behavior does not happen locally.

fdejean:
Talk to the bug to Abel, he suggest it was the screen size of the size, add an args to the CI fixed it.

app/src/main/java/com/android/feedme/MainActivity.kt
```diff
@@ -43,6 +45,7 @@ class MainActivity : ComponentActivity() {
               // Navigation host for the app
               val navController = rememberNavController()
               val navigationActions = NavigationActions(navController)
+              val profileViewModel = ProfileViewModel()
```
ohamz, line 48: A better way to instantiate the profileViewModel would be as follows:
`val profileViewModel = viewModel<ProfileViewModel>()`

Let me know what you think, I remember doing it the current way in the bootcamp led to some bugs.

app/src/main/java/com/android/feedme/ui/profile/EditProfileScreen.kt
```diff
@@ -0,0 +1,211 @@
...
+ * This function sets up the Scaffold layout which includes a top bar, bottom navigation, and the
+ * content for editing the profile, handled by [EditProfileContent].
+ *
+ * @param navigationActions A set of actions to handle navigation events.
+ * @param profileViewModel The ViewModel that handles fetching and updating the user's profile data.
+ */
+@Composable
+fun EditProfileScreen(
+    navigationActions: NavigationActions,
+    profileViewModel: ProfileViewModel = ProfileViewModel()
```
ohamz, line 46: Same note as before, I think it would be better to instatiate the viewModel as follows:
`viewModel<ProfileViewModel>()`

ohamz (changes requested):
I suggested changing the Profile view model instatiations to `viewModel<ProfileViewModel>()` but it raises errors and tells me change the parameter type to `Lazy<ProfileViewModel>` instead of `ProfileViewModel` when passing it by argument.

Have you tried it ?

app/src/main/java/com/android/feedme/MainActivity.kt
fdejean: I do not understand why. How is this better ? Why do want to declare the type explicitly ? @ohamz

ohamz: I can't explain it well, but this is the response to ChatGPT when I asked which is better : 



- *Using ProfileViewModel() directly instantiates the ViewModel. This approach is straightforward but comes with certain limitations:*

*- Lifecycle Awareness:* The ViewModel is not lifecycle-aware. It won't survive configuration changes like screen rotations. If the activity or fragment is recreated, a new instance of the ViewModel is created, which can lead to loss of data.
*- Resource Management:* It does not handle resources or ongoing processes well. If the ViewModel is managing a process or network call, those could be abruptly stopped or restarted, leading to inefficiencies or errors.


- *Using viewModel<NewViewModel>(), typically seen in Android development with Jetpack's ViewModel architecture, has several advantages:* 

*- Lifecycle Management:* The ViewModel is lifecycle-aware. It survives configuration changes such as screen rotations. This means data isn't lost when the activity or fragment is recreated.
*- Resource Efficiency:* Handles resources more efficiently. Ongoing processes managed by the ViewModel are not interrupted unnecessarily.
*- Scoped to Lifecycle:* It is scoped to the lifecycle of the activity or fragment, ensuring it is cleaned up when the lifecycle ends, preventing memory leaks.

fdejean: Okay, but I dont think the profileViewModel will have life Cycle at all.
The user may frequently check their profile or other person, hence to re-create one is costly.
So I create the instance in mainActivity as never end so it survive configuration.
I don't think the Resource Management make any sense too since it does not die, there should not be any interruption.

ohamz (approved):
Perfect ! 
It passes all test and the requested changes were made and approved.

This PR can be merged.

Issue #67, opened 2024-04-12
Follower/Following UI


0marwa:
Code the Figma implementation with static content inside of the app

PR #75, opened 2024-04-16
68 viewmodel for followerfollowing page

fdejean:
The idea is to have a friend page (follower/following UI) that is dynamic and based on the friends stored on the firebase database

### DIFF
--- a/app/src/main/java/com/android/feedme/MainActivity.kt
+++ b/app/src/main/java/com/android/feedme/MainActivity.kt
@@ -22,7 +22,7 @@ import com.android.feedme.ui.auth.LoginScreen
 import com.android.feedme.ui.camera.CameraScreen
 import com.android.feedme.ui.navigation.NavigationActions
 import com.android.feedme.ui.navigation.Route
-import com.android.feedme.ui.profile.FollowersScreen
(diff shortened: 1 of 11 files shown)


fdejean:
Waiting for #67 to be merge into main to put this PR in review

app/src/main/java/com/android/feedme/MainActivity.kt
```diff
@@ -43,6 +45,7 @@ class MainActivity : ComponentActivity() {
               // Navigation host for the app
               val navController = rememberNavController()
               val navigationActions = NavigationActions(navController)
+              val profileViewModel = ProfileViewModel()
```
ohamz, line 48: The instantiation has to be modified as the new way, 

Done in another PR and will be fixed when merging.

ohamz (approved):
I like the navigation system used, can we do this for any subscreen ? If yes I will adapt my Camera route and navigation implementation.

Otherwise, approved !

fdejean:
> I like the navigation system used, can we do this for any subscreen ? If yes I will adapt my Camera route and navigation implementation.

We can but it has to be useful, as for example the recipe with their Id, see the Notion guide on how to implement!

PR #78, opened 2024-04-18
Script for scraper of foodnetwork

xrsula:
This script scrapes all of the recipes on the website foodnetwork. It follows this procedure:
- It starts by the base html of the recipes (a URL that leads to the recipes classified by alphabetical order) and scrapes all recipes within the page,
- It continues by automatically going page by page (alphabetical order) and scrapes all recipes on each page.
- The information found in the recipes are stored in 4 different types of .json files: one that links every title of recipe to a unique ID, one for each recipe that contains all relevant information of said recipe, one that joins all ingredients used across all recipes, and finally one that joins all categories mentioned across all recipes.
- The information is simply dumped in the .json files, and for now needs further processing / cleanup. (Can be modified to include processing within this script)
- There is also an additional timer, that waits 5 to 10 seconds between each html request in order to avoid being IP banned.

/!\ Need to still add the nutritional values of each recipe

Update of 25/04/2024:
We will not be using this script for copyright issues, but it has been developed already and will stay on the git repo in case we have to use it. It works, but we have found open source alternatives to fill our database with !


fdejean (approved):
Impressive! 
It will be greatly useful to fill the database with real data !
The next step to implement this in the app is quite clear : parse the json such that it match the database json and that seem quite straightforward.
Good job!

mche3kek (approved):
Nice job with the script ! Concise and does a good job with the .json formatting for easier treatment and integration with the database !

PR #79, opened 2024-04-18
Fixed a bug with ProfileViewModel

fdejean:
Description : upon sign-in, the default profile is display instead of the database profile
Fix : add a listener


ohamz (approved):
Nice catch of the problem.
Fixes approved.

PR #80, opened 2024-04-18
Fixing the tests

ohamz:
Added a functional mock for the Login that handles a successful sign in that I added to the end-to-end UserFlow test.
Arrange the tests of the individual screens (they won't handle navigation, it will be handled in the end-to-end tests).


JeannedeM (approved):
Looked at it very quickly, but seems good to me :)

PR #81, opened 2024-04-18
57 UI for displaying the full recipe page

JeannedeM:
The goal is to code the UI to display the recipe on full page, like this :
322051060-daebf2fd-6954-4751-9ae6-3d238bec18a9
Acceptance criteria : 

- Top bar displayed with the name of the recipe + the save icon
- Image of the recipe displayed
- General informations displayed (**Warning** : filter about the recipe difficulty is not yet implemented in our data model because it depends on the recipes we will scrap. Thus I will not display this information for now).
- Ingredients for the recipes displayed in a list fashion.
- (**Warning** : ustensils are not yet implemented in our data model because it depends on the recipes we will scrap. Thus I will not display this information for now)
- Recipe's steps are displayed.


ohamz (approved):
Approved. Very good coverage as well !

PR #83, opened 2024-04-18
Fix an issue with 0 followers

fdejean:
Was missing a check if the list was null before querying the database leading to a crash


ohamz (approved):
Bug fixed :)

PR #84, opened 2024-04-19
Navigation to FullRecipe

ohamz:
Added the navigation to FullRecipeScreen when clicking on a recipe in the LandingPage.

I added a temporary recipe for the moment in order to show it in the demo, will be removed once the RecipeViewModel is created.


mche3kek (approved):
Looks good to me

PR #91, opened 2024-04-19
Updating the landing page to match the exact Figma description 

0marwa:
I updated the landing page to match the Figma as asked in my previous PR. The tests are implemented as well.
One detail is the uploading dates of the Recipes are not available for me to show so I will modify the data class `Recipe` and then add those to the Recipe Cards.


fdejean:
Hello!
If you wish to modify the data model, be sure to update RecipeRepository, too
I think it's better to create another issue. In the meantime, you can just hard code the date :)

JeannedeM (changes requested):
The result is great ! Niceeee job ! Here are some changes that could be great imo : 

Put the Search bar out of the RecipeDisplay function maybe ? it do not seems very logical to put that inside a composable function that iterates over all recipes to generate recipe card as it's not a recipe card itself. But i don't know, i might not have understand well :)
(If you do this modification it would rougly be extract lines 112 to 148 and put it in an other function with its specific documentation).
Otherwise the documentation and the TODO are great Good job 💯

ohamz:
I really like the new design, a picture below for reference, 
image

@JeannedeM as for the SearchBar, I think the current implementation is fine. If you take a look, the recipe cards are displayed within a scrollable LazyColumn, but outside there is a Column composable that englobes the SearchBar and the LazyColumn, do you see it ?

0marwa:
Thanks for the feedbacks! 

While keeping the search bar as is is perfectly fine (upon discussion with the rest of the group), we may need to reuse it in future code. So, to avoid duplicating code, I'll simply move it to a separate component. :)

JeannedeM (approved):

PR #92, opened 2024-04-20
Implement NavHost as a Nested Graph

ohamz:
Modify the MainActivity to support nested navigation.

Fixed the Route object to include only the 6 main routes, other 'routes' were added to the Screen object.

Waiting for the RecipeViewModel to be implemented to fix navigation to FullRecipeDisplay.


ohamz:
Will be adding end-to-end tests to improve MainActivity coverage

fdejean (approved):
Perfect! 
Nothing to add, expect that I am really happy to have a clean navigation before that we add to many nested pages.

PR #93, opened 2024-04-21
RecipeViewModel implementation

ohamz:
Implement the RecipeViewModel.

This view model will be created when generating the landing page then linked to the full recipe display when navigating to it. This way we don't have to fetch the recipe from the database (because we already have it).


ohamz:
Only 1 line uncovered in RecipeViewModel, I think it's well tested enough.
The rest of uncovered lines is from other files.

app/src/main/java/com/android/feedme/MainActivity.kt
```diff
@@ -59,11 +62,16 @@ class MainActivity : ComponentActivity() {
...
-                    }
+                  composable(Screen.HOME) {
+                    // Create a shared view model for Recipe
+                    val recipeViewModel = viewModel<RecipeViewModel>()
+                    LandingPage(navigationActions, recipeViewModel)
+                  }
+                  composable(Screen.RECIPE) {
+                    // Link the shared view model to the composable
+                    val navBackStackEntry = navController.getBackStackEntry(Screen.HOME)
+                    val recipeViewModel = viewModel<RecipeViewModel>(navBackStackEntry)
```
fdejean, line 73: Duplicate of line 67, they are the same instance and will die if we leave `Screen.HOME`.
However, another is an issue, as the `RecipeFullDisplay` may be access from the Profile, do we re-create another `RecipeViewModel` for only that Route or should we share it with` Screen.HOME` (i.e a single `RecipeViewModel` for the whole app) ?

fdejean, line 73: Another problem may be, to be sure that the `RecipeViewModel` is set to the correct recipe we want.
One fix for that would be to kept the Id navigation and compare it upon creation to fetch the `Recipe` if needed.
However, this scenario should never happen as we always should be able to set the `Recipe` we want to display before hand. Do you agree ?

fdejean (changes requested):
The current code fulfill, it's purpose : No unnecessary request to Firestore is made to navigate to a recipe. I checked using log that it was the case. 
However, there is a small fix needed to avoid a duplicated line and it should be perfect.
Good job!

app/src/main/java/com/android/feedme/MainActivity.kt
ohamz: We are sure that we'll have the correct recipe. When navigating to `RecipeFullDisplay` from `LandingPage`, I'm selecting the `Recipe` we click on before navigating to the screen using `selectRecipe(recipe)` on the shared `ViewModel` (on line 132 of `LandingScreen.kt`). 
Does this answer your doubt ?

ohamz: We discussed privately and came to the conclusion that this case will be dealt later when implemented.

fdejean (approved):

PR #94, opened 2024-04-22
View other user profiles

FacerOfGod:
Able to view a follower's profile
- Being able to navigate to the _follower/following_  list and click on
- Follow button appears and changes from follow  <-> following when clicked
- When viewing your profile you should still be able to edit your profile

### DIFF
--- a/app/src/main/java/com/android/feedme/MainActivity.kt
+++ b/app/src/main/java/com/android/feedme/MainActivity.kt
@@ -79,7 +79,11 @@ class MainActivity : ComponentActivity() {
                 }
 
                 navigation(startDestination = Screen.PROFILE, route = Route.PROFILE) {
-                  composable(Screen.PROFILE) { ProfileScreen(navigationActions, profileViewModel) }
+                  composable(Screen.PROFILE) { backStackEntry ->
+                    backStackEntry.arguments?.getString("profileId")?.let {
+                      ProfileScreen(navigationActions, profileViewModel, it)
+                    } ?: run { ProfileScreen(navigationActions, profileViewModel) }
+                  }
                   composable(Screen.EDIT_PROFILE) {
                     EditProfileScreen(navigationActions, profileViewModel)
                   }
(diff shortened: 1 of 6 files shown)


app/src/main/java/com/android/feedme/MainActivity.kt
fdejean, line 1: This is a bit inefficient, as we already have the profile object we wish to display. Hence for every profile view, we will actually do two fetches. 
I think Omar approach is better, he stored the Recipe directly in the ViewModel before navigating to the screen.

app/src/main/java/com/android/feedme/model/viewmodel/ProfileViewModel.kt
fdejean, line 1: I think not storing the Profile of the login user to be inefficient as this profile will be important for the app in general and must not be fetched all the time we need it. Hence having a variable loginProfile make sense,

fdejean (changes requested):
There is an inefficiency in how we fetches profiles, it's important to minimize the number of request we do to firestore,
Otherwise, the code is clean and is minimized, you did great at adapting previous code to implement the new features !

app/src/main/java/com/android/feedme/MainActivity.kt
ohamz: @FacerOfGod You can see how I implemented it in these files `MainActivity.kt`, `LandingScreen.kt` and `RecipeFullDisplay.kt` (in `home` package).
Don't hesitate to ask questions if there's something unclear.

FacerOfGod:
Perfect, with your advice in mind , I moved every data manipulation from the `ProfileScreen `in the `ProfileViewModel` like it should be, so only UI is present in `ProfileScreen`. `ProfileViewModel` now has the abilty to set a viewingUser and it always stores the currentUser profile.

fdejean (approved):

fdejean:
The change are correctly implemented! Nice work

PR #105, opened 2024-04-26
script modified and adapted for new website

xrsula:
We were able to find an open source website that hosts more than 3000 recipes, with varying levels of details.

This jupyter script enables us to scrape the contents of the website in order to collect all data of all recipes, and store them in 3 different .json files:
- One that contains all categories that appear at least once in the website (not filtered, need to remove irrelevant tags),
- One that contains all ingredients that appear at least once in the website (not filtered, need to remove irrelevant quantities and descriptions)
- One that contains all the recipes with their details (not filtered, some need to have some text fixed)

The website we used is https://en.wikibooks.org/wiki/Category:Recipes, and their disclaimer enables us to use the contents of their website.


scraping/script.ipynb
```diff
@@ -19,25 +19,103 @@
...
+    "    # Extract the href attributes, focusing on individual recipe pages\n",
+    "    recipe_links = []\n",
+    "    for link in soup.find_all('a', href=True):\n",
+    "        href = link['href']\n",
+    "        # Look for links that are likely to be individual recipes\n",
+    "        if href.startswith('/wiki/Cookbook:') and not \"Category:\" in href:\n",
+    "            full_url = base_url + href\n",
+    "            recipe_links.append(full_url)\n",
+    "\n",
+    "    return recipe_links[2:-1]\n",
```
ohamz, line 83: Maybe just add a little comment specifying why first 2 and last elements of the return list aren't considered.

scraping/script.ipynb
```diff
@@ -90,35 +230,37 @@
...
     "    print(\"All URLs have been processed.\")\n",
     "    \n",
     "    # Writing the index file and ingredient/category accumulations\n",
-    "    write_json(recipe_index, \"recipe_index.json\")\n",
+    "    write_json(recipe_index, \"recipes.json\")\n",
     "    write_json(list(all_ingredients), \"all_ingredients.json\")\n",
     "    write_json(list(all_categories), \"all_categories.json\")\n",
     "\n",
-    "# main() TODO: UNCOMMENT TO RUN"
+    "main() # TODO: UNCOMMENT TO RUN"
```
ohamz, line 263: Running this script will print the results on the output console right?

ohamz (changes requested):
Script implementation seems very clean, it was already tested so this confirms that it works !

scraping/script.ipynb
xrsula: Yes it will ! Since I am using a jupyter notebook to run the code it will print out the results in an output cell directly under the cell of my code.

scraping/script.ipynb
xrsula: Good point I will :)

scraping/script.ipynb
ohamz: Ok great!

ohamz (approved):
Approved!
We now have to figure out how to store these Recipes in the Firestore database.

PR #112, opened 2024-04-28
SignOut button

ohamz:
Implemented the SignOut button in SettingsScreen and added a main test for display.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,6 +18,7 @@
 *.zip
 *.tar.gz
 *.rar
+*.jks
 
 # virtual machine crash logs, see http://www.java.com/en/download/help/error_hotspot.xml
 hs_err_pid*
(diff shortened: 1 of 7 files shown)


ohamz:
Should I add a test for onClick functionality of SignOut button ? 
I can try but I think it will cause trouble as when we tested the sign in (when it used to fail alot on CI so we removed it).

xrsula (changes requested):
Everything looks good for me ! It would be nice to specify the TODO for the top bar in SettingsScreen.kt to see what you plan on doing for it ! a small comment to describe the contents of it is sufficient

ohamz:
> Everything looks good for me ! It would be nice to specify the TODO for the top bar in SettingsScreen.kt to see what you plan on doing for it ! a small comment to describe the contents of it is sufficient

The UI for the settings screen is left for another issue, that's why I minimalized my part so that this PR focuses on the backend of the button. But I agree it's a good idea to add some comments in order to mark it clear.

xrsula (approved):
All good :)

PR #113, opened 2024-04-28
Profile Navigation Bug

fdejean:
Found a bug from the PR : https://github.com/feedme-swent-epfl/feedme-android/pull/94
Found when implementing : https://github.com/feedme-swent-epfl/feedme-android/pull/95

Scenario : go to profile, navigate to one of your following, open their profile, look for their following.
Problem : will show the default profile.


fdejean:
Solution : Wrong variable updated when setting a new "viewing profile".
I am surprised this was not found when doing testing.

FacerOfGod (approved):
Nice catch! This saves us from further trouble down the line.

ohamz:
> Solution : Wrong variable updated when setting a new "viewing profile". I am surprised this was not found when doing testing. Could be that the database was fill with the same follower/following value, leading to the bug to be undiscovered

I think I found this bug once when I was navigating through the app. Next time I encounter a similar bug I'll take note to not forget it and address the person that implemented the feature.

PR #116, opened 2024-04-30
Viewmodel for Manual Input

FacerOfGod:
Goal of this PR : Develop the backend functionality for the manual ingredient input screen.
- [x]  Implement InputViewModel to replicate the existing functionality.
- [x]   Maintain the list of inputs and display even after navigation transitions.
- [x]  For the future : prepare a function to return the list of ingredients when fully completed.
- [x]  For the future : allow to easily display a list of ingrendients after a scanning.


JeannedeM (changes requested):
Great Job, your code for this view model is clean and concise ! Here are some details to modify in my opinion : 

- You are mixing the name of Ingredient and IngredientMetadata i think ? If i am not mistaken user is inputing IngredientMetadata not Ingredient in this part of the app. I think should adpat documentation and variables name accordingly :)
- I think you should also change the name of the function setUpdate() line 67 to something like updateListElement() or something else. I found the current name very confusing 👀 .
- I am also not sure to understand how you can completly delete an ingredient from list using the updating function or the adding function. But maybe it's not a useful functionnality to have idk.

Otherwise i think it's all good :D

FacerOfGod:
> * I am also not sure to understand how you can completly delete an ingredient from list using the updating function or the adding function. But maybe it's not a useful functionnality to have idk.

I maintain a variable called `totalIngredientEntriesDisplayed`, which signifies the number of entries to be displayed. Usually, it equals the size of `listOfIngredientMetadatas`, plus one extra element representing an empty entry.

I manage `IngredientMetadatas` with states, and based on their state transitions, I either remove the element with their index (indexing is crucial for proper deletion) or add a new null element if the previous one was utilized. 

The function `addToList` is intended for adding a list of detected ingredients (from text, barcode, or object recognition). However, I refrain from using this function during manual ingredient addition to ensure a smoother user experience. Transitioning users between composables disrupts their writing flow.

I think with the time and experience I got, this is the best I have got. Perhaps in the future, and I hope, that we'll discover an better solution :)

JeannedeM (approved):
Thanks for the clarification and the modification !

PR #121, opened 2024-05-02
108 UI for barcode recognition

JeannedeM:
Create the UI for the barcode recognition functionnality. Requirements :

- [x] - A button is dipslayed on the right side of the camera screen
- [x] - The button icon represent clearly the scanning barcode functionnality
- [x] - All button are uniform in term of style


app/src/main/java/com/android/feedme/ui/camera/Camera.kt
```diff
@@ -157,16 +160,37 @@ fun CameraScreen(navigationActions: NavigationActions) {
...
+                      modifier =
+                          Modifier.size(56.dp)
+                              .background(CameraButtonsBackground, shape = CircleShape)
+                              .padding(10.dp)
+                              .testTag("MLBarcodeButton")) {
+                        Icon(
+                            painter = barcodeScannerPainter,
+                            contentDescription = "Barcode Scanner",
+                            modifier = Modifier.size(25.dp),
+                            tint = Color.Black)
```
fdejean, line 191: I would be better to use the Color theme of the App instead of `import androidx.compose.ui.graphics.Color`.

fdejean (changes requested):
Only a small issue, otherwise all good!

app/src/main/java/com/android/feedme/ui/camera/Camera.kt
JeannedeM: Thanks you are right !

fdejean (approved):

PR #122, opened 2024-05-02
107 implement ml kit for bar code recognition

JeannedeM:
Requirements : 

- [x] Implement dependencies of google Ml kit for barcode scanning
- [x] User can take a photo of a barcode to scan it
- [x] After scanning : raw value of the barcode is displayed
- [x] Relevent error message is shown in case of failure


xrsula (approved):
Overall very nice job ! You were able to implement the barcode scanning, and the code is clean and well documented. Next step is to link the value of the barcode with the ingredients, good job :)

PR #129, opened 2024-05-02
UI and implementation of Welcome page and editing of profile at creation

xrsula:
This PR deals with the welcome page. 

It implements the UI of the Welcome Page, which pops up when a user first logs in and is not present in the database of profiles, and leads them to the editing page where they can edit the components of their profile


0marwa (approved):
Well documented code, and well done for the tests! :) Approved

PR #135, opened 2024-05-03
Store profile picture in Firestore 

mche3kek:
Let the user choose a profile picture which will be synced in our database for accross all devices


app/src/androidTest/java/com/android/feedme/test/ui/EditProfileTest.kt
```diff
@@ -146,6 +146,14 @@ class EditProfileTest {
...
       editPicture.performClick()
+      // TODO inject a mock image picker
+      /*
+            profileViewModel.currentUserId = "ID_DEFAULT"
+            profileViewModel.updateProfilePicture(
+                profileViewModel,
+                Uri.parse(
+                    "https://encrypted-tbn0.gstatic.com/images?q=tbn:ANd9GcT0ouuiiXjqpSyZLcW7yXsrGl9T-FxgUQJMNSaoF3JUitDQ_t-oVrJjbCCQkfYmdGYJoeM&usqp=CAU"))
+      */
     }
```
ohamz, line 157: Why is this commented ? Can it be removed ?

app/src/main/java/com/android/feedme/model/viewmodel/GalleryViewModel.kt
```diff
@@ -21,15 +21,15 @@ import kotlinx.coroutines.flow.MutableStateFlow
 
 class GalleryViewModel : ViewModel() {
 
-  val _bitmaps = MutableStateFlow<List<Bitmap>>(emptyList())
```
ohamz, line 24: Should this val be private ?

ohamz (approved):
This seemed like a really complex PR. Well done !

Tested and approved.

app/src/main/java/com/android/feedme/model/viewmodel/GalleryViewModel.kt
mche3kek: Nope, as mentioned in the other PR it is used in another class

app/src/androidTest/java/com/android/feedme/test/ui/EditProfileTest.kt
mche3kek: I left it in case we want to test it, it would require to mock a user with write ability which i did not have time to do

PR #130, opened 2024-05-02
109 implement api of open food repo

JeannedeM:
Goals : 

- [x] User can scan a barcode by taking a photo with camera
- [x] Information about product is retrieved from on online api with an http request
- [x] User can read basic informations about the product (ie. the product name)****
- [x] If the scanner cannot detect a barcode in the photo, a relevant error message is displayed
- [x] If no photo has been taken nothing is displayed****

Future improvements in a nextPR : 

- Product information's are displayed in a user friendly way (not the raw json string)
- Error messages are more adapted and precise depending on the situation
- Code for ML operation written in the Camera.kt file will be moved to the CameraViewModel.kt file for more logical organisation.

PR #137, opened 2024-05-03
Fix CI problem

mche3kek:
CI is malfunctionning for Jeanne and I ( https://github.com/feedme-swent-epfl/feedme-android/pull/130 and https://github.com/feedme-swent-epfl/feedme-android/pull/135 ), probably due lack of disk space from error messages.
This PR seeks to fix that.


mche3kek:
Seems like the CI is working properly now, I am checking if the issue disappeared from issue #130 just by deleting older artifacts or not, in which case I will probably make that job run in a regular manner rather than at each CI run

mche3kek:
We have another issue that seems to be propagated in the CI, maybe freeing the cache could help overcome these kind of issues ?

mche3kek:
Turns out I am stupid and misunderstood something, I'm gonna push back the cache

mche3kek:
I think it's gonna be more useful to test it out directly on my branch, I am gonna close this pull request for that reason

Issue #146, opened 2024-05-03
ViewModel for Input of recipe Step Creation


fdejean:
Create the logic behind the input of creation of step to be able to delete, add a step.
The UI will be made in #141

PR #147, opened 2024-05-03
141 UI for recipe step creation

fdejean:
Display a list of Step

- [x] Match the figma
- [x] Added a colapse feature
- [x] Added error warning for empty title and description

### DIFF
--- a/app/src/main/java/com/android/feedme/ui/component/RecipeStepInput.kt
+++ b/app/src/main/java/com/android/feedme/ui/component/RecipeStepInput.kt
@@ -1,0 +1,110 @@
+package com.android.feedme.ui.component
+
+import androidx.compose.foundation.clickable
+import androidx.compose.foundation.layout.*
+import androidx.compose.foundation.lazy.LazyColumn
+import androidx.compose.material.icons.Icons
+import androidx.compose.material.icons.filled.DeleteForever
+import androidx.compose.material.icons.filled.ExpandLess
+import androidx.compose.material.icons.filled.ExpandMore
+import androidx.compose.material3.*
+import androidx.compose.runtime.*
+import androidx.compose.ui.Alignment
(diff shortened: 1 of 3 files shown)


0marwa:
I'll leave a preview here for reference (the background is transparent for some reason?) 
Screenshot 2024-05-05 at 12 47 50

0marwa (approved):
Good job having the description field shown when we expand the title, as the description is optional :) 
Other than that, nothing to signal!

app/src/main/java/com/android/feedme/ui/component/RecipeStepInput.kt
```diff
@@ -0,0 +1,110 @@
...
+          },
+          onDeleteStep = {
+            // stepViewModel.deleteStep(index)
+          })
+    }
+  }
+}
+
+@Composable
+fun StepInput(step: Step, onStepChanged: (Step) -> Unit, onDeleteStep: () -> Unit) {
```
JeannedeM, line 41: Hello, great UI ! Thank you ! Could you just add documentation for the function here please :)

JeannedeM (changes requested):

app/src/main/java/com/android/feedme/ui/component/RecipeStepInput.kt
fdejean: Hey, yes, I will add some! 
Thanks

fdejean:
@0marwa @JeannedeM thanks for the reivew,
Request changed have been implemented,

Here is the UI with a white background
Note that the step number depend on the Step object, and since it use the same step, the step number does not change.
That behavior will change in #146 

Screenshot_20240505_144810_feedmeApp

JeannedeM (approved):

PR #148, opened 2024-05-03
Create viewModel at the correct place

fdejean:
Bug ; Upon new sign-in, nothing is shown in landing page
Fix : ViewModel was created before the sign-in, hence was not able to fetch database upon init


JeannedeM (approved):
Perfect, thank you !

PR #149, opened 2024-05-03
Finalize M2 submission

mche3kek:
Addition of SCRUM Board and demo video to res folder, bug fixes for gallery pictures and username in profile creation, details on APK workflow file


fdejean (approved):
Nice!

PR #151, opened 2024-05-05
146 viewmodel for input of recipe step creation

fdejean:
## User Action

- [x] Able to create a step
- [x] Able to delete a step
- [x] Upon deletion of a step, re-order the step number

## UI Update

- [x] Add a Title to the List of Step 

## Output

- [x] ViewModel must make accessible the List of Step variable for the creation of recipe


JeannedeM (approved):
Hey, great work ! Everything seems good to me ! Maybe we will do Small UI updates in the future, but that's only details.

PR #153, opened 2024-05-05
Searchbar logic

ohamz:
### Implement search logic when interacting with the search bar from the Home page.

- [x] Add fetching method to RecipeRepository.
- [x] Add fetching method to ProfileRepository.
- [x] Adapt the SearchBar to support the logic.

~~[ ] Adapt the UI to support the searching of profiles and recipes.~~ As mentioned below this will be implemented in a future PR.

### Update:

- [x] Find out why it's not working and fix it.
- [x] Improve searching technique

### Final tasks:
- [x] TEST

### DIFF
--- a/app/src/main/java/com/android/feedme/model/viewmodel/HomeViewModel.kt
+++ b/app/src/main/java/com/android/feedme/model/viewmodel/HomeViewModel.kt
@@ -1,6 +1,5 @@
 package com.android.feedme.model.viewmodel
 
-import android.util.Log
 import androidx.lifecycle.ViewModel
 import androidx.lifecycle.viewModelScope
 import com.android.feedme.model.data.Profile
@@ -93,9 +92,6 @@ class HomeViewModel : ViewModel() {
    * @param query: the query to search for in the recipes
    */
   fun searchRecipes(query: String) {
-    Log.d("HomeViewModel", "Filtered recipes: ${_filteredRecipes.value}")
-    Log.d("HomeViewModel", "Recommended recipes: ${_recommendedRecipes.value}")
-    Log.d("HomeViewModel", "Recipes: ${recipes.value}")
     viewModelScope.launch {
       recipeRepository.getFilteredRecipes(
           query,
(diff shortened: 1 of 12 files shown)


ohamz:
Logic is working now, but needs a bit of reworking:
- Search substring of target query instead of exact full query
- Testing is needed 
- UI is needed to be adjusted

ohamz:
This PR is already too long, I think UI will be done in a different PR (for Profiles and Recipes in the HOME page).

ohamz:
Testing took a lot of time, almost 8 hours.
I added 2 unit tests for searchRecipes and searchProfiles in HomeViewModelTest.

For now when running the app we can only visualize the recipes being filtered, since I haven't implemented the full UI for searching yet (will be implemented in next PR).
Later we will be able to visually see profiles and recipes being filtered.


Another note, the filtering algorithm is as follows:
If the input query is a prefix of a certain recipe in the database, this recipe will be shown.

I made it as a prefix and not a substring because we will have almost 4k Recipes, and finding a substring will require us to fetch all 4k recipes then filter them out one by one. The implemented method is far more efficient.

app/src/main/java/com/android/feedme/model/viewmodel/HomeViewModel.kt
```diff
@@ -0,0 +1,160 @@
...
+          })
+    }
+  }
+
+  /**
+   * A function that fetches the recipes given their Ids
+   *
+   * @param ids: the unique IDs of the profiles we want to fetch
+   */
+  /*fun fetchRecipes(ids: List<String>) {
```
JeannedeM, line 65: What is exactly this commented code ?

ohamz: This method was present before I started working on the view model.
I believe it's a method that will be modified in a later issue in order to fetch multiple recipes at the same time (in 1 query). It will be used when we will fetch the base recipes in Home page (it's written below for the recommendation algorithm).

JeannedeM: ok thanks for the clarification ! Otherwise really good job !! Thank youuu :D

JeannedeM (approved):

PR #155, opened 2024-05-08
143 clean up of mlkit implementation

JeannedeM:
Requirements : 

- calls to various ML function in Camera.kt file have been moved to the camera view model.
- wrapping functions in the camera view model allow easy access & execution of ML functionnality from the outside.

Improvements for future PR : 

- Fix the multiples error messages.
- Create adapted display composable for each ML application.
- Maybe create a specific display composable for error messages ? (this one could be common to all ML application).
- Improve the tests to have the required coverage (this will be done when error messages are fixed).
- An issue is already prepared for that : parsing of jsonfile and extrcat relevant informations.


app/src/main/java/com/android/feedme/ml/Barcode.kt
```diff
@@ -50,21 +50,20 @@ fun barcodeScanner(
...
  * @param barcodeNB The barcode number for which the product name needs to be extracted.
- * @param onSuccess Callback function invoked when the product name extraction is successful. It
- *   takes a single parameter, [productName], representing the extracted product name.
+ * @param onSuccess Callback function invoked when the product name extraction is successful.
  * @param onFailure Callback function invoked when an exception occurs during the extraction
- *   process. It takes a single parameter, [exception], representing the exception that occurred.
+ *   process.
  * @throws Exception if an error occurs during the extraction process.
  */
 suspend fun extractProductNameFromBarcode(
```
FacerOfGod, line 58: In this function in terms of good pratice just keep the exception as an input for the onFailure function, even if for now it is useless. We might need it down the line.

app/src/main/java/com/android/feedme/ml/Barcode.kt
```diff
@@ -50,21 +50,20 @@ fun barcodeScanner(
...
     barcodeNB: String,
-    onSuccess: (String) -> Unit,
-    onFailure: (Exception) -> Unit
+    onSuccess: (String) -> Unit = {},
+    onFailure: () -> Unit = {}
 ) {
   try {
     onSuccess(httpRequestBarcode(HttpMethod.GET, barcodeNB, "fields=product_name"))
   } catch (e: Exception) {
     println(e.message)
```
FacerOfGod, line 66: You are using println, I suggest using Log.e(...) for better clarity.

FacerOfGod (changes requested):
Nice job !! It makes `Camera.kt` a lot easier to work with. I think that there are just a few changes to make for good partices mesures that I have commented below. Other than that great job :D

FacerOfGod (approved):
Well done :D

app/src/main/java/com/android/feedme/ml/Barcode.kt
JeannedeM: Got it !

PR #124, opened 2024-05-02
Added Label to Scrapped Ingredients

fdejean:

PR #156, opened 2024-05-09
Script to add ingredients to Firebase DB done and works

xrsula:
This python script helps to add the ingredients to the database of firebase.

The ingredients are stored by Unique IDs generated by Firebase, and have the attributes name (for the name of the ingredient), and whether it is vegetarian or vegan. Eventual added attributes can be for dairy content and gluten conetnt, but will be decided later on.

The script was run on my laptop and the database was correctly filled in !


mche3kek (changes requested):
The script seems good and the ingredients are correctly uploaded. However, there are some mistakes with the labeling of some (Nuoc Mam is vegan and vegeterian for example, c.f screenshots below for other example
image

xrsula:
> The script seems good and the ingredients are correctly uploaded. However, there are some mistakes with the labeling of some (Nuoc Mam is vegan and vegeterian for example, c.f screenshots below for other example image

This seems to come from a previous implementation done by @fdejean in the PR #124 so I would love to have some input from him !

fdejean:
Hey, 
Yes, some miss labeling will happen, especially for complex words.
However, I feel this will not be such an issue as a recipe will use "all the ingredients" to know if a recipe is vegan or vege. So, the logical function is more likely to not consider a recipe to be vege/vegan. 
If this later is proved to be an issue, we should refilter the database to only contain single words and re-label them

mche3kek (approved):
I'll trust Florian on this as the small error in labeling won't affect much the result, I'll approve of this PR

PR #160, opened 2024-05-10
158 error handling mlkit

JeannedeM:
Goal : 

- Fix error messages to make them consistent and informative for user.

### DIFF
--- a/app/src/main/java/com/android/feedme/ml/TextRecognition.kt
+++ b/app/src/main/java/com/android/feedme/ml/TextRecognition.kt
@@ -1,6 +1,7 @@
 package com.android.feedme.ml
 
 import android.graphics.Bitmap
+import android.util.Log
 import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.padding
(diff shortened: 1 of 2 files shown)


mche3kek (commented):
Everything looks good to me, is it voluntary to have these errors only show in the Logcat instead of displaying a message to the user indicating that it failed?

JeannedeM:
> Everything looks good to me, is it voluntary to have these errors only show in the Logcat instead of displaying a message to the user indicating that it failed?

The message in Logcat is just here for debugging pupose, a message is indeed displayed to the user each time an error occurs. this is just not visivle here, as the message is written in the camera view model and thrown when the onFailure function is called by textExtraction :)

mche3kek (approved):
> > Everything looks good to me, is it voluntary to have these errors only show in the Logcat instead of displaying a message to the user indicating that it failed?
> 
> The message in Logcat is just here for debugging pupose, a message is indeed displayed to the user each time an error occurs. this is just not visivle here, as the message is written in the camera view model and thrown when the onFailure function is called by textExtraction :)

Perfect, thanks so much for the clarification ! Everything looks great you can go ahead and merge it !!

PR #164, opened 2024-05-11
Navigation from full recipe display AND RecipeCard to Profile

ohamz:
Adapted the username implementation and added navigation to the user's Profile.

If a Recipe does not have a user creator, or this user's id could not be fetched, the username Text won't be displayed, you can see below both possibilities.

Screenshot 2024-05-11 233623

Screenshot 2024-05-11 232431

### Update: 
I noticed that the username is also visible from the RecipeCard in the Home Page AND Search Page, so I also added the navigation from both.

### DIFF
--- a/app/src/androidTest/java/com/android/feedme/test/LandingTest.kt
+++ b/app/src/androidTest/java/com/android/feedme/test/LandingTest.kt
@@ -11,6 +11,7 @@ import com.android.feedme.model.data.Recipe
 import com.android.feedme.model.data.RecipeRepository
 import com.android.feedme.model.data.Step
 import com.android.feedme.model.viewmodel.HomeViewModel
+import com.android.feedme.model.viewmodel.ProfileViewModel
 import com.android.feedme.model.viewmodel.RecipeViewModel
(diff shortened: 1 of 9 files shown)


ohamz:
This PR is ready to be merged, there are only 4 uncovered lines that's why line coverage for this PR is below 80%.

ohamz:
Ready for review :)

xrsula (approved):
Good job ! This code is well written and does the job nicely of navigating to a profile (if it exists) from a recipe. It is also well documented and easily understandable !

A small remark for improvement would be to document the more complicated tests, since sometimes it can be quite difficult to follow what the test is supposed to do and how it is set up !

fdejean (approved):
Test in the app, work great.
The navigation to a profile from a thumbnail work well but the UI is a bit confusing as the difference between the end of an description and the username is not enough

ohamz:
> the UI is a bit confusing
I didn't make UI changes as this PR was only meant to add the navigation logic. However I did refactor a bit by creating a new composable for RecipeCard to make it clearer and reusable.

PR #169, opened 2024-05-14
UI implementation for the saved recipe screen

xrsula:
This PR deals with the implementation of the basic UI for the saved recipes screen, which shows for now "You did not save any recipes yet!" since the backend has not been implemented yet 
A next PR will deal with the linking of the profile with the screen

cf. the picture below, which shows what the saved screen currently displays (and will display when a user has no saved recipes)
Screenshot 2024-05-14 at 23 11 06


0marwa (approved):
Well documented, thanks for including screenshots of the final result! (:

Issue #97, opened 2024-04-25
ViewModel for the comments


0marwa:
With this we should be able to fetch comments from the database and display them.

0marwa:
I'm keeping this issue but lowering its priority as it's not really relevant for this EPIC

0marwa:
We realised this ViewModel is very similar (identical) to that of creating a new comment on a recipe. And @fdejean has already started on it so it'll be done in the issue #174

PR #175, opened 2024-05-15
Add Logic/Navigation to Create a comment

fdejean:
## UI Update
- [x] Added Pop-up logic to view the CreateComment UI 
- [x] Show the CreareScreen upon the FBA clicked
- [x] Close the CreateScreen upon the "submit" or "cancel" button is clicked

## Action Create a comment
- [x] Set the comment in the database and get it corresponding ID
- [x] Add the submited comment to the database (/comments/commentsId)

## ViewModel Update
- [x] Create a comment View Model, utilize CommentRepository to set a comment to the database

## Warning
This PR does not implement the update of commentId list of profile and recipe. New back-end function are required for it, and will be done at a later date in #177

### DIFF
--- a/app/src/main/java/com/android/feedme/model/data/CommentRepository.kt
+++ b/app/src/main/java/com/android/feedme/model/data/CommentRepository.kt
@@ -37,7 +37,7 @@ class CommentRepository(private val db: FirebaseFirestore) {
    */
   fun addComment(comment: Comment, onSuccess: () -> Unit, onFailure: (Exception) -> Unit) {
     db.collection("comments")
-        .document(comment.authorId)
+        .document(comment.commentId)
         .set(comment)
         .addOnSuccessListener { onSuccess() }
         .addOnFailureListener { exception -> onFailure(exception) }
--- a/app/src/main/java/com/android/feedme/model/data/CommentRepository.kt
+++ b/app/src/main/java/com/android/feedme/model/data/CommentRepository.kt
@@ -29,21 +29,19 @@ class CommentRepository(private val db: FirebaseFirestore) {
   }
 
   /**
-   * Adds a comment to Firestore.
-   * Will OVERRIDE the comment Id and get a new from firestore.
+   * Adds a comment to Firestore. Will OVERRIDE the comment Id and get a new from firestore.
    *
    * @param comment The Comment object to be added.
    * @param onSuccess Callback invoked on successful addition of the comment.
    * @param onFailure Callback invoked on failure to add the comment, with an exception.
    */
   fun addComment(comment: Comment, onSuccess: () -> Unit, onFailure: (Exception) -> Unit) {
-      val newDocRef = db.collection("comments").document()
-      comment.commentId = newDocRef.id // Assign the generated ID to the comment
-      newDocRef.set(comment)
-          .addOnSuccessListener { onSuccess() }
-          .addOnFailureListener { exception ->
-              onFailure(exception)
-          }
+    val newDocRef = db.collection("comments").document()
+    comment.commentId = newDocRef.id // Assign the generated ID to the comment
+    newDocRef
+        .set(comment)
+        .addOnSuccessListener { onSuccess() }
+        .addOnFailureListener { exception -> onFailure(exception) }
   }
 
   /**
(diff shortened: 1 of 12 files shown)


app/src/main/java/com/android/feedme/model/data/CommentRepository.kt
```diff
@@ -29,15 +29,16 @@ class CommentRepository(private val db: FirebaseFirestore) {
...
+   * Adds a comment to Firestore. Will OVERRIDE the comment Id and get a new from firestore.
    *
    * @param comment The Comment object to be added.
    * @param onSuccess Callback invoked on successful addition of the comment.
    * @param onFailure Callback invoked on failure to add the comment, with an exception.
    */
   fun addComment(comment: Comment, onSuccess: () -> Unit, onFailure: (Exception) -> Unit) {
-    db.collection("comments")
-        .document(comment.authorId)
+    val newDocRef = db.collection("comments").document()
```
ohamz, line 39: Are we creating a collection for comments ? 
How will we link it to the corresponding recipe and profile, is it only through the userId and recipeId ?

app/src/main/java/com/android/feedme/model/viewmodel/CommentViewModel.kt
```diff
@@ -0,0 +1,52 @@
...
+    _comment.value = comment
+  }
+
+  /**
+   * A function that add a comment in the database Overwrites the comment object given to assign it
+   * a new commentId
+   *
+   * @param comment: the comment to set in the database
+   */
+  fun addComment(comment: Comment, onSuccess: () -> Unit) {
```
ohamz, line 38: Good way to add the comments to the database, consistent with what we've done in previous methods.

ohamz (changes requested):
Great implementation, the code is clean and well documented.

Only one problem I got when testing is that I wasn't able to scroll down in the FullRecipeDisplay to see the 'Dialog' after publishing a comment. Is that normal ?

app/src/main/java/com/android/feedme/model/data/CommentRepository.kt
fdejean: > Are we creating a collection for comments ?

No, here is the definition of `document()`  :
Returns a DocumentReference pointing to a new document with an auto-generated ID within this collection.

> How will we link it to the corresponding recipe and profile, is it only through the userId and recipeId ?

 If you look at the model, you will see that that for a recipe and a profile, each have a list containing commentId.

fdejean:
> Only one problem I got when testing is that I wasn't able to scroll down in the FullRecipeDisplay to see the 'Dialog' after publishing a comment. Is that normal ?

I am not sure what you are talking about ? `Dialog` refer to the pop-up to show the create a comment UI

app/src/main/java/com/android/feedme/model/data/CommentRepository.kt
fdejean: The logic may be confusing,  first we create a comment with a dummy Id, then we ask firestore to give us a new id, and finally we update our comment with this id.

ohamz: Okay I see, I thought the a single comment was accessible 3 times in the database (user, recipe and comment), but since in the user and the recipe we only have a list of 'commentId' it should be fine :)

ohamz:
> > Only one problem I got when testing is that I wasn't able to scroll down in the FullRecipeDisplay to see the 'Dialog' after publishing a comment. Is that normal ?
> 
> I am not sure what you are talking about ? `Dialog` refer to the pop-up to show the create a comment UI

So the UI for displaying the comments on a recipe is still not implemented ?

fdejean:
> So the UI for displaying the comments on a recipe is still not implemented ?

No, this is not the content of this PR, I believed the PR #97 was supposed to do it.

ohamz:
> No, this is not the content of this PR, I believed the PR #97 was supposed to do it.

Yes I thought it was implemented in a previous PR that's why I found it weird that it doesn't work.

ohamz (approved):
Approved !
We will have to add the UI for displaying comments on a recipe in a future PR.

PR #180, opened 2024-05-16
144 parsing json files for ingredient input through barcode

JeannedeM:
Goals :

- [x]     Parse Json string properly.
- [x]     Modify what informations are displayed to user after barcode scanning.
- [x]     Make those informations more user friendly based on the json parsing.
- [x]     Scanned product are added to the ingredient list.

### DIFF
--- a/app/src/main/java/com/android/feedme/ml/Barcode.kt
+++ b/app/src/main/java/com/android/feedme/ml/Barcode.kt
@@ -64,8 +64,8 @@ suspend fun extractProductInfoFromBarcode(
     onFailure: (Exception) -> Unit = {}
 ) {
   try {
-      val result = httpRequestBarcode(HttpMethod.GET, barcodeNB, "fields=product_name")
-      println(result)
+    val result = httpRequestBarcode(HttpMethod.GET, barcodeNB, "fields=product_name")
+    println(result)
     onSuccess(parseJsonString(httpRequestBarcode(HttpMethod.GET, barcodeNB, "fields=product_name")))
   } catch (e: Exception) {
     e.message?.let { Log.i(e.message, it) }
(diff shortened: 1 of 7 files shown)


JeannedeM:
Once again we decided to not have full coverage on this function. We cannot add barcode images in the CI Emulator hence it's almost impossible to test the function.....

app/src/main/java/com/android/feedme/model/viewmodel/CameraViewModel.kt
mche3kek, line 1: An ingredient will always have a default metadata ? Why so ?

app/src/main/java/com/android/feedme/ml/Barcode.kt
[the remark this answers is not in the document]
JeannedeM: Good question ! it's an information that i receive when i make a request to the Api, basically it tells me if the API has information about a product. Sometimes the API also has partial informations only.

app/src/main/java/com/android/feedme/model/viewmodel/CameraViewModel.kt
JeannedeM: It's only temporary until Sylvain implements the functionnality to add unknown Ingredient to the data base in his next PR.

app/src/main/java/com/android/feedme/ml/Barcode.kt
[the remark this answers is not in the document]
mche3kek: Okay, so it is useful for deciding if we output an error or not, thanks for the clarificattion !

app/src/main/java/com/android/feedme/model/viewmodel/CameraViewModel.kt
mche3kek: Oh I see, thanks for the precision

mche3kek (approved):
Everything looks good to me :)

PR #182, opened 2024-05-16
BUG FIX : Crash on Start

fdejean:
Weird bug that is not related to recent push but due to the database update.
Landing Page fetch some default recipe but they have been deleted.
The view model was wrongly coded to threw and error if a recipe is not found.
Replaced the threw with a Logging instead.

### DIFF
--- a/app/src/main/java/com/android/feedme/model/data/RecipeRepository.kt
+++ b/app/src/main/java/com/android/feedme/model/data/RecipeRepository.kt
@@ -102,6 +102,7 @@ class RecipeRepository(private val db: FirebaseFirestore) {
     db.collection(collectionPath)
         .whereGreaterThanOrEqualTo("title", query)
         .whereLessThan("title", query + "\uf8ff")
+        .limit(10)
         .get()
         .addOnSuccessListener {
           it.documents.map { recipeMap ->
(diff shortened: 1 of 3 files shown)


mche3kek:
You need to do a KTFMT format :)

fdejean:
The real bug is that the firestore document path name has been changed. Hence, the onFailure is called.

This occured when @0marwa @xrsula uploaded the reciped to the database.
Can you modify the database to have the correct path name ? 
@0marwa @xrsula thanks you :)

fdejean:
Another bug was found :
We exceed our quota extremely quickly with the new filled scraped recipes.
I think this is due to the search bar, as the query get "all the matching criteria" and does not limit the size.
Hence a query with "a" will get all the recipes containing an "a". 
Fix : added a limit close the query of the search bar

ohamz (approved):
Well done catching these bugs, and implementing safe measures for the moment until we fix them !

fdejean:
Well, the fix of the search was needed, but the thing that exceed our quota is simply 
- Using the firestore webpage make queries. Do not manually look for a recipe and scroll. 
Just loading the firestore /recipes make about 2k request....

PR #187, opened 2024-05-16
Suggest/Add Ingredient to Database in InputViewModel

FacerOfGod:
Goal : Recommend ingredients as users input them and enable the addition of ingredients not yet in the database.

- [x] Integrate "Add Ingredient" UI feature
- [x] Develop the backend functionality for adding ingredients
- [x] Implement backend search for ingredients
- [x] Optimize database requests for efficiency
- [x] Elements detected by the camera should appear in green
- [ ] Thoroughly test all functionalities

~Ingredients added throught the camera will be considered as unknown for now. This will change in a next PR.~


app/src/main/java/com/android/feedme/model/data/Ingredient.kt
```diff
@@ -1,3 +1,8 @@
 package com.android.feedme.model.data
 
-data class Ingredient(val name: String, val type: String, val id: String)
+data class Ingredient(
+    val name: String,
+    var id: String = "NO_ID",
```
JeannedeM, line 5: I think it could be better to encode the type of the ingredient as `val type: IngredientType = VEGAN` with `IngredientType` being an enum that we can extend. Otherwise future changes are gonna be really complicated :(

app/src/main/java/com/android/feedme/ml/TextRecognition.kt
```diff
@@ -220,7 +219,7 @@ fun parseResponse(responseBody: String?, forIngredientFound: (IngredientMetaData
       val unitString = ingredientObject.optString("unit", "")
       val unit = getMeasureUnitFromString(unitString)
       forIngredientFound(
-          IngredientMetaData(quantity, unit, Ingredient(ingredient, "DEFAULT_TYPE", "DEFAULT_ID")))
+          IngredientMetaData(quantity, unit, Ingredient(ingredient, "NO_ID", false, false)))
```
JeannedeM, line 222: I saw that "`NO_ID"` was the default value at Ingredient creation, but here why : 

- don't we give a real idea to ingredient ?
- if we don't, why are we initialising it to `"NO_ID"` even though it's already the deafult value at creation ?

app/src/main/java/com/android/feedme/model/viewmodel/CameraViewModel.kt
```diff
@@ -216,13 +217,11 @@ class CameraViewModel : ViewModel() {
...
                   { productInfo ->
                     if (productInfo != null) {
-                      // TODO How can i create new ingredients correctly or check if they already
-                      // exist ? => Sylvain PR ?
                       updateIngredientList(
                           IngredientMetaData(
                               0.0,
                               MeasureUnit.NONE,
-                              Ingredient(productInfo.productName, "Default", "DefaultID")))
+                              Ingredient(productInfo.productName, "NO_ID", false, false)))
```
JeannedeM, line 224: Same here for the `"NO_ID"`

app/src/main/java/com/android/feedme/model/viewmodel/CameraViewModel.kt
```diff
@@ -287,10 +286,31 @@ class CameraViewModel : ViewModel() {
    * @param ing The ingredient metadata to update the list with.
    */
   fun updateIngredientList(ing: IngredientMetaData) {
+    if (ing.ingredient.id != "TEST_ID") {
```
JeannedeM, line 289: what is `"TEST_ID"` here ? was it for test purpose ?

JeannedeM (changes requested):
Hey, this PR represent really a tremendous amount of work !! Good job 🔥 🔥 🔥 I added some comments about changes I think would be relevant :)

app/src/main/java/com/android/feedme/model/viewmodel/CameraViewModel.kt
FacerOfGod: It's the easiest way to prevent the uses of `IngredientsRepository.instance.getExactFilteredIngredients()` when testing.

app/src/main/java/com/android/feedme/ml/TextRecognition.kt
FacerOfGod: This is a problem with how our model works, elements that are detected need to be transformed into `Ingredient` types before knowing or not  if their Ingredient exist in Firebase. So this allows me to distinguish between Ingredients that are actually in the database or not. I agree that this isn't optimal, but for now it works.

app/src/main/java/com/android/feedme/model/viewmodel/CameraViewModel.kt
JeannedeM: Ok i see !

app/src/main/java/com/android/feedme/model/data/Ingredient.kt
FacerOfGod: I agree with you. However, I thing this should be in another PR because refactoring Ingredient demands a lot of energy and it isn't the main point of this PR.  I think we should also have the whole group agree on how to do it before refactoring. What do you think?

app/src/main/java/com/android/feedme/ml/TextRecognition.kt
FacerOfGod: It's just the way I code, we could remove it if necessary. I find it easier to code like this :)

JeannedeM: Ok no problem, it's datails anyway :)

app/src/main/java/com/android/feedme/model/viewmodel/CameraViewModel.kt
JeannedeM: Solved in conversation above.

JeannedeM:
Also while testing i found a bug in the UI : after clicking on "Add Ingredient", the drop down menu never desappear... Could you try to fix that issue please :) 
image

app/src/main/java/com/android/feedme/model/data/Ingredient.kt
JeannedeM: Ok, is think that because we are gonna refactor the database tommorow, it's useless to do it now. Let's keep that for another PR.

JeannedeM (approved):
Once again thank you for your work. Everything works perfectly now ! The result is amazing !

PR #188, opened 2024-05-16
Database infinite query fix

ohamz:
I modified the getFilteredRecipes and getFilteredProfiles methods in RecipeRepository and ProfileRepository respectively.

A limit of 10 was added for each call if these methods, and at the end of the screen I added a 'Load more' button to fetch 10 more Recipes/Profile (if available) when clicked.


fdejean (approved):
Neat fix ! 
One suggestion to keep in mind : 
Add a de-bouncing on the search bar such that if a user is still typing, it only do the query when is done and not at every input. This will already remove around 10xn queries for each use. (n  = number of char typed).

PR #189, opened 2024-05-17
Python script to upload recipes to database

0marwa:
This python script was ran on my laptop and allowed us to upload the recipes to the FireStore Database (: 
Small fixes following the script : changing the `recipe` path to allow us to fetch from the right `collection` in the Database


scraping/updated_recipes.json
mche3kek, line 1: Is this the updated file with the description filled by ChatGPT or not ? If not could we add that instead ?

scraping/fill_db_recipes.ipynb
mche3kek, line 1: The script looks good to me, only thing I want to point out is that the paths are customized to your laptop, wouldn't it be better to have a relative path relevant to our project structure instead so that anyone can run the script ?

mche3kek (changes requested):

xrsula:
The script needs some updating because of the new way we want to store the recipes in the database. Since I was the one that dealt with the creation of the recipe.json and the upload of the ingredients in the database, I will commit a small fix that will deal with that

xrsula (changes requested):
This script looks great ! Apart from Mamoun's remark, I would say that you will now need to upload to the database the new recipe list that contains the curated ingredients. This thus requires a small change to the script :)

scraping/updated_recipes.json
0marwa: It's been added below!

mche3kek (approved):
Everything looks good to me !

xrsula (approved):
Looks good :) Thanks for the changes !

PR #206, opened 2024-05-18
Bug fix step input issue

fdejean:
Due to re-computation of the step, the keyboard retract preventing to correctly input the step.
Add the re-computation upon losing focus instead that on every step


mche3kek (approved):
Everything looks good to me, works flawlessly !

mche3kek (approved):

PR #210, opened 2024-05-19
UI fix search bar

JeannedeM:
- [x]     Remove the 'Filter' Icon
- [x]     Put the 'Search' Icon at the left of the search bar when not clicked (instead of 'Filter')
- [x]     Deal with the decomposition of the search bar when clicked
- [x]     Make the querry text visible


ohamz (approved):
Nice fix for the search bar ! We can now see the text we're inputting.

JeannedeM:
Thanks for fixing the tests !

JeannedeM (commented):
Perfect and efficient fix, well done !

PR #216, opened 2024-05-19
Implement Offline mode

ohamz:
Implement Login and navigation through the app when no internet connection.

### Tasks : 

- [x] User can still log in without internet connection.
- [x] Update navigation to disable functionnality only avaible when having internet connection (ie. create the user flow in offline mode).

### Update: 
- [x] Adapt existing tests
- [x] TESTING time (offline)

## For reviewer(s):
Try to traverse all screens and functionalities of the app Online AND Offline in order to try and detect bugs or crashes. 
If you are not already signed in, you won't be able to sign in Offline.


app/src/main/java/com/android/feedme/model/data/RecipeRepository.kt
```diff
@@ -38,10 +41,22 @@ class RecipeRepository(private val db: FirebaseFirestore) {
...
+  fun addRecipe(
+      recipe: Recipe,
+      context: Context = FirebaseFirestore.getInstance().app.applicationContext,
+      onSuccess: () -> Unit,
+      onFailure: (Exception) -> Unit
+  ) {
+    // Check if the user is offline
+    if (!isNetworkAvailable(context)) {
+      Log.d("addRecipe", "Offline mode: Cannot add recipe")
+      return
```
fdejean, line 57: `isNetworkAvailable` is a clever implementation, but why are you not calling onFailure before returning ?

fdejean (commented):
Excellent job!
The app let me login while being offline, then upon deleting my cache and phone data, the app did not let me logging again as expected :)

However a side effect I saw during testing : Multiple `Toast` stack.
Hence while navigating, I had the pop-up open for the whole time I was using the app...

app/src/main/java/com/android/feedme/model/data/RecipeRepository.kt
ohamz: If I call onFailure it would throw exceptions in the logs, that's not what we want since this is an intended functionality.
However, I agree that the Toasts are stacking up, that's because I'm displaying it whenever the method isNetworkAvailable, is called and returns false, maybe I should display the Toast somewhere else.

fdejean: No, onFailure does not throw an exception, it's a function defined by the caller.
Hence it can be used to do pattern matching on the exception and display information in a more friendly way.

Not calling onFailure could break the app for feature that need a answer, such as generatedRecipe as we are waiting for an answer with a "loading",  yet if nothing happen, the app may be stuck in this loading state  forever....

However, I don't think such a loading has been implemented yet, that's why I did not require change.

ohamz:
> However a side effect I saw during testing : Multiple `Toast` stack.

This should be fixed now :) 
Tell me if it's better now or not.

fdejean:
> > However a side effect I saw during testing : Multiple `Toast` stack.
> 
> This should be fixed now :) Tell me if it's better now or not.

There is no more spam, and navigation does not create a toast anymore!
I will re-review once you respond my initial review comment

app/src/main/java/com/android/feedme/model/data/RecipeRepository.kt
ohamz: So you want me to call the method's `onFailure` each time the network connection is checked and unavailable ?

fdejean: > So you want me to call the method's `onFailure` each time the network connection is checked and unavailable ?

It would be the ideal but I think the current implementation work for now. 
It's the best-way without spending too much time checking all the caller ^^

fdejean (approved):
Perfection!

ohamz:
Got 81% total line coverage :)

PR #217, opened 2024-05-19
Generated Recipes Screen

ohamz:
Implement the UI for displaying the recipes recommended in a new subscreen.

- [x] Implement UI for generated recipes screen.
- [x] Link frontend with backend implementation.
- [x]  Test all.
- [x] Get required coverage.


ohamz:
I have to wait for the recommendation system backend to be implemented before completing this PR.

ohamz:
Ready for review :)


@fdejean Could you check if the backend works please, because I have a feeling it doesn't (unless I forgot something when linking it to the UI).

fdejean:
> Ready for review :)
> 
> @fdejean Could you check if the backend works please, because I have a feeling it doesn't (unless I forgot something when linking it to the UI).

Yes, there was in issue with database missing an IDs field. 
I fixed it work now!

app/src/main/java/com/android/feedme/model/viewmodel/GenerateViewModel.kt
```diff
@@ -0,0 +1,88 @@
...
+   * A function that fetches the generated recipes
+   *
+   * @param ingredientIds: the unique IDs of the recipes we want to fetch
+   * @param profile: the profile of the user
+   */
+  fun fetchGeneratedRecipes(ingredientIds: List<String>, profile: Profile) {
+    _ingredientsIds = ingredientIds
+    _profile = profile
+
+    if (_isStrict.value) {
```
fdejean, line 39: The logic is reversed, by default, the toggle button is on "strict" and it's seem the value is true as when sliding to "extra" is does the strict search.

app/src/main/java/com/android/feedme/model/data/RecipeRepository.kt
```diff
@@ -325,21 +326,31 @@ class RecipeRepository(private val db: FirebaseFirestore) {
...
   ) {
     if (ingredientIds.isEmpty()) {
       onFailure(IllegalArgumentException())
       return
     }
-    db.collection(collectionPath)
-        .whereArrayContainsAny("ingredientIds", ingredientIds)
+    // Create a query to fetch the top rated recipes
+    var queryRef =
+        db.collection(collectionPath).whereArrayContainsAny("ingredientIds", ingredientIds).limit(6)
```
fdejean, line 346: No, you cannot do that, here you limit the recipes to 6. Hence the ranking will happen only on those 6 recipes and not all the recipe that match criteria. Either increase this limit or do a filtering on the highest ranked recipes first.
The best way will be to utilize SQL firebase or BigQuerry but it was decided that local filtering will work for now

fdejean (changes requested):
Nice work, but some details need to changed for the feature to work as excepted (c.f my comment)

app/src/main/java/com/android/feedme/model/data/RecipeRepository.kt
ohamz: Hmmm I see, so where do you think can I add this `.limit()` call ? (I did these changes to be able to add the Load more button)

app/src/main/java/com/android/feedme/model/viewmodel/GenerateViewModel.kt
ohamz: Ohhh my bad, nice catch !

app/src/main/java/com/android/feedme/model/data/RecipeRepository.kt
fdejean: The process is done in "one go" 
You can't really pause it and get a coherent result. 
(Would be the case if it was a not local suggestion)

Hence, there is no load more feature. The output is a complete list, for the UI you could use a LazyList and it should display only the requested element.

ohamz: I wanted to implement the Load more button functionality, but I think I will just display all fetched recipes on after the other.

fdejean (approved):
Nice

0marwa (approved):
Good job spotting and fixing the DB bugs, no remarks on my end; the code is well-documented so well done!

PR #218, opened 2024-05-19
Displaying best recipes on Landing Page

ohamz:
Fetch the best rated recipes from the database in order to display them on the landing page below the search bar.


app/src/main/java/com/android/feedme/ui/component/LoadMoreButton.kt
mche3kek, line 1: Nice job with the button ! Do you think you could use derived paddings from the screen length instead of fixed lengths to avoid UI bugs across different sized devices ?

app/src/test/java/com/android/feedme/ui/HomeViewModelTest.kt
mche3kek, line 1: Nice testing !

app/src/main/java/com/android/feedme/ui/SavedScreen.kt
mche3kek, line 1: Why was the saved screen deleted ? Is it integrated somewhere else ?

app/src/main/java/com/android/feedme/model/data/RecipeRepository.kt
```diff
@@ -66,14 +67,15 @@ class RecipeRepository(private val db: FirebaseFirestore) {
   }
 
   /**
-   * Fetch all the recipes of the given List of Ids
+   * Fetches the top rated recipes from Firestore. The recipes are ordered by their rating in
```
mche3kek, line 70: This method would be for saved recipes instead of top rated ones

app/src/main/java/com/android/feedme/model/data/RecipeRepository.kt
```diff
@@ -93,10 +95,50 @@ class RecipeRepository(private val db: FirebaseFirestore) {
...
+   * descending order.
+   *
+   * @param lastRecipe The last recipe fetched in the previous query. If null, fetches the first
+   *   page of recipes.
+   * @param onSuccess A callback function invoked with the list of recipes on success.
+   * @param onFailure A callback function invoked on failure to fetch the recipes, with an
+   *   exception.
+   */
+  fun getRatedRecipes(
+      lastRecipe: DocumentSnapshot?,
```
mche3kek, line 109: Nice addition to avoid fetching the six same recipes continually

mche3kek (changes requested):
Nice code and I really love the button ! I have two points to raise though :

1) is it normal to already have 6 recipes (apart from lasagna and pasta) already displayed ?

2) when clicking on any recipe that is not Pasta or Lasagna, the app crashes with an Invalid document reference instead of displaying the recipe, is it normal (I guess not but we never know) ?

app/src/main/java/com/android/feedme/model/data/RecipeRepository.kt
ohamz: Nice catch, I'll fix it.

app/src/main/java/com/android/feedme/ui/component/LoadMoreButton.kt
ohamz: For the padding it's not a problem, since I added enough padding for the button to be visible and not touching with the bottom screen.
The only possible issue would be fixed width, but it is still centered with enough space on the borders so it shouldn't cause problem.

ohamz:
> 1. is it normal to already have 6 recipes (apart from lasagna and pasta) already displayed ?

I have 6 recipes in total, including the lasagna and pasta, it turns to 12 when I click on "Load more" button


</pre>


> 2. when clicking on any recipe that is not Pasta or Lasagna, the app crashes with an Invalid document reference instead of displaying the recipe, is it normal (I guess not but we never know) ?

Yes this is an issue that we have to fix, it's when clicking on the scrapped recipes in the database like you said, but from any screen not just from HOME (SEARCH and SAVED). This is an important issue that has to be fixed in another PR.

mche3kek (approved):
Everything looks good to me ! Only concern on my part would be to open a PR for the bug if it is not already the case. Great job !

PR #219, opened 2024-05-20
Fix the UI of following button in Profile Screen

xrsula:
In order to fix the UI of the following button in Profile Screen, I added additional padding to main UI components in order to make the screen fit better any type of phone screen

### DIFF
--- a/app/src/main/java/com/android/feedme/ui/profile/ProfileScreen.kt
+++ b/app/src/main/java/com/android/feedme/ui/profile/ProfileScreen.kt
@@ -12,7 +12,6 @@ import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.height
 import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.layout.width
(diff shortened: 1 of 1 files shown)


ohamz:
@JeannedeM can you test it on your device if it works ? If I remember the problem was when launching the app on a physical device, on the emulator it's fine.

image

JeannedeM:
> @JeannedeM can you test it on your device if it works ? If I remember the problem was when launching the app on a physical device, on the emulator it's fine.
> image

I tried on my phone and it was still the same problem :(
image

JeannedeM:
I just push something that fixes the roblem on my phone, but it would be great of @xrsula  and @ohamz  could test it, to be sure it's not a mess on their phones....

ohamz:
> I just push something that fixes the problem on my phone, but it would be great of @xrsula and @ohamz could test it, to be sure it's not a mess on their phones....

It still works on the emulator so I think the fix is good !

image

ohamz (approved):
Bug fixed !

PR #225, opened 2024-05-20
224 bug  ingredients not displayed in full recipe screen

fdejean:
Scenario : Upon visiting a full recipe, no ingredients appear... Yet they are well uploaded in the database....
What could be the bug?

### DIFF
--- a/app/src/main/java/com/android/feedme/model/data/IngredientsRepository.kt
+++ b/app/src/main/java/com/android/feedme/model/data/IngredientsRepository.kt
@@ -2,6 +2,7 @@ package com.android.feedme.model.data
 
 import android.util.Log
 import com.google.firebase.firestore.DocumentSnapshot
+import com.google.firebase.firestore.FieldPath
 import com.google.firebase.firestore.FirebaseFirestore
 import com.google.firebase.firestore.Query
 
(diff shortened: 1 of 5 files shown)


fdejean:
The bug was in the back-end : fetching of the `Ingredient` Class.
Was returing `MetaDataIngredient` instead of `Ingredient`
Also simplified fetching of a recipe to avoid fetching each ingredient as every info is already stored directly in the recipe with the newly migrated database.

JeannedeM (commented):
Nice job for fixing this bug ! Do you think it's possible to reach 80% coverage or not ?

app/src/main/java/com/android/feedme/model/data/IngredientsRepository.kt
```diff
@@ -101,39 +102,25 @@ class IngredientsRepository(val db: FirebaseFirestore) {
...
+        .whereIn(FieldPath.documentId(), ingredientIds)
+        .get()
+        .addOnSuccessListener { querySnapshot ->
+          val ingredients =
+              querySnapshot.documents.mapNotNull { documentSnapshot ->
+                documentSnapshot.toObject(Ingredient::class.java)
+              }
+          onSuccess(ingredients)
+        }
+        .addOnFailureListener { exception -> onFailure(exception) }
```
ohamz, line 123: Good fix, so you're getting the `Ingredients` directly instead of `IngredientMetaData`.

app/src/test/java/com/android/feedme/model/IngredientsRepositoryTest.kt
```diff
@@ -181,8 +184,13 @@ class IngredientsRepositoryTest {
...
 
-    // Setup mocks to simulate a failure for any fetch attempt
-    `when`(mockDocumentReference.get()).thenReturn(Tasks.forException(exception))
+    // Setup mocks to simulate a failure for the whereIn query
+    `when`(mockFirestore.collection("ingredients")).thenReturn(mockCollectionReference)
+    `when`(
+            mockCollectionReference.whereIn(
+                Mockito.eq(FieldPath.documentId()), Mockito.eq(ingredientIds)))
+        .thenReturn(mockQuery)
+    `when`(mockQuery.get()).thenReturn(Tasks.forException(exception))
```
ohamz, line 194: The tests are well adapted to the changes.

ohamz (approved):
The bug is tested and is fixed now, the only uncovered lines are due to an enum, which wasn't covered in the tests for some weird reason.
It won't affect the overall coverage since it's only a couple of lines.

Approved!

app/src/main/java/com/android/feedme/model/data/IngredientsRepository.kt
fdejean: Yes

fdejean:
> Nice job for fixing this bug ! Do you think it's possible to reach 80% coverage or not ?

No, I adapted all the test, none where removed.
I had to add some new line to handle the unit conversion correctly, as we stored them in string in the database.
It seem sonar cloud don't like nested function and hence does not consider the whole function to be covered.
Passing 20minutes to add a test to check to get those 10missing line of coverage does not seems relevant now ...

PR #227, opened 2024-05-21
Enhance Search Bar functionality

ohamz:
Upon searching a recipe, the app crash when the recipe is clicked on (Actually this problem arises when clicking on any of the database recipes, for example lasagna1 and pasta1 work). 
And also the searching logic has to be modified.

- [x] Crash Free
- [x] Change the backend for better searching experience
- [x] All tests are adapted and pass

### Update: 
So I modified the search bar logic, instead of searching for an exact prefix of the title of a recipe, now you can search the words of the title.

For example if I have a recipe with title "Delicious Apple Pie":
- The old logic would display this recipe if the query is "D", "De", ..., "Delicious", "Delicious A", ..., "Delicious Appl", ..., "Delicious Apple Pie".
- The new logic will display this recipe if the query is the following words: "Delicious", "Apple", "Pie", and it's case insensitive so it can be "DELICIOUS", "apple", "PiE", etc ...

I did this by adding a new field to the recipes, `searchItems`, that contains the words of the title in an array of lowercase strings (["delicious", "apple", "pie"] for the example above), and before I compare the input query to check if it's withing this list, I convert it to lower case.


ohamz:
The only files that I modified are: 
- `SearchViewModel.kt`
- `SearchViewModelTest.kt`

All the rest was 'modified' due to rebasing and resolving conflicts with main.

app/src/main/java/com/android/feedme/model/data/RecipeRepository.kt
```diff
@@ -107,32 +134,53 @@ class RecipeRepository(private val db: FirebaseFirestore) {
...
       onFailure: (Exception) -> Unit
   ) {
+    // Prefix search query
     var queryRef =
         db.collection(collectionPath)
-            .whereGreaterThanOrEqualTo("title", query)
-            .whereLessThan("title", query + "\uf8ff")
-            .limit(10) // Limit the number of documents fetched
+            .whereArrayContainsAny("searchItems", listOf(query.lowercase(Locale.ROOT)))
+            .limit(6) // Limit the number of documents fetched
```
mche3kek, line 141: Why did we choose a limit of 6 ?

mche3kek: Also it would be nice to explain what the query does exactly with a comment

app/src/main/java/com/android/feedme/model/data/RecipeRepository.kt
```diff
@@ -107,32 +134,53 @@ class RecipeRepository(private val db: FirebaseFirestore) {
...
         .addOnFailureListener { onFailure(it) }
   }
 
+  /**
+   * A helper function that adds the recipes to the list of recipes
+   *
+   * @param snapshot: the snapshot of the query
+   * @param onSuccess: the callback function to be called on success
+   * @param onFailure: the callback function to be called on failure
+   */
```
mche3kek, line 159: Nice way to avoid high code complexity ! Would love to have a bit more comments as the code is obscure, I have a hard time understanding what it does exactly

mche3kek (changes requested):
Some functions lack documentation, otherwise good enhancement !

For the search logic change I think it is a shame that we don't support partial word queries, as with only a certain substring (e.g "del") displaying all recipes which titles contain that substring.

The recipe is displayed twice, and there is  a non-functional load more button when testing on emulator :
image

ohamz:
> For the search logic change I think it is a shame that we don't support partial word queries, as with only a certain substring (e.g "del") displaying all recipes which titles contain that substring.

It's easier said than done ... I did some research and this implementation was the best solution between effectiveness and efficiency. Searching for a substring will require us to fetch ALL the recipes from the database and compare them one by one with the query.

 > The recipe is displayed twice, and there is a non-functional load more button when testing on emulator

The recipe is displayed twice because there are duplicate recipes in the database.
As for the load button, it isn't working because there are no more recipes to fetch (maybe a future PR would be to remove this button when it's the case).
If you try with "chicken" you can try and see if the button works.

app/src/main/java/com/android/feedme/model/data/RecipeRepository.kt
ohamz: Will add some comments 👍

app/src/main/java/com/android/feedme/model/data/RecipeRepository.kt
ohamz: Recipe cards take enough space in the UI so a limit of 6 is more reasonable than 10 (before the user can press on load more)

mche3kek (approved):

Issue #212, opened 2024-05-19
Implement photo upload to comments


JeannedeM:
This issue's aim is to implement the photo uploading to comments, so when a user comments on a recipe, they can add photos from their gallery to illustrate the results they got. This will likely reuse some of the code done by @mche3kek for the profile pic photo upload from gallery

PR #228, opened 2024-05-21
223 UI fix  refactor the comment input on recipe

JeannedeM:
Fix the input of comment such that :

 

- [x]    Text in "Publish" button is written on one line
- [x]     You can see the number you are entering for the "star rating"
- [x]     The star icon is Yellow (same yellow as for the other stars)
- [x]     the text in "Delete" button -> becomes "Cancel" button (we are not deleting a comment because it's not even created).
- [x]     Make the whole thing a bit more fancy ?
- [x]     Ensure that the picture upload from this [PR ](https://github.com/feedme-swent-epfl/feedme-android/issues/212)goes well.

Further informations : 

- I copy paste the exact composable of the future profile picture, hence it should not cause problem when we implement it in this [PR ](https://github.com/feedme-swent-epfl/feedme-android/issues/212).
- The stars and half stars are clickable and directly translate into a rating over 5 (step over 0.5).
- I also changed the floating button icon and added something more representative of a comment.

TODO in other PRs: 

1. Implement real profile picture.
2. Link the UI to the creation of a real comment.
3. Do test on the **functionnality** of this page (currently, only the layout is being tested).


JeannedeM:
Result :)
image

xrsula (approved):
Everything looks good to me ! I like how you changed the way we rate from a string format to a double (decimal number) format !
Something that needs to be taken into consideration is that the rating is not being saved in the Database, so another PR has to be made for that :)
Overall, the code is well documented, and I can easily understand your modifications, well done !

0marwa (approved):
Good job updating the UI, I agree that it's better than the one agreed upon before :) The rating method is better and regrouping the text related fields under the picture just allows for a clearer filling method. A couple remarks for what still need to be done : 
- store the rating in the database (to be added in a future PR)
- implementing the photo upload process (another PR is already open for this one)
- update the Figma prototype to match this new outline 

Other than that, good job on the documentation and clarity of both your PR and code :)

PR #229, opened 2024-05-22
Fix : title on one line and corrected save screen

JeannedeM:
When the page title is more than one word, its displayed on two lines and not one.
Tasks :
- [x]    Modify the UI of the Top Bar component
- [x]     Verify that this change apply well to all pages of the app.

331888159-b4482298-9958-40b5-9427-43b4dd117537

Additionnaly, do this small UI fix :

- [x]   The top bar title for the saved recipe screen is "FeedMe" and should be "Saved Recipe".


ohamz (approved):
Good fix well done !

PR #231, opened 2024-05-22
Ingredient generation popup UI implementation

xrsula:
In this PR, I implemented UI of the popup for recipe generation. The text gives information to the user of the two ways to generate their recipes: 
- Strict : the user only generates recipes with those ingredients
- Extra : the user generated recipes that may contain additional ingredients compared to the ingredients inputted.

The popup has this appearance :

### DIFF
--- a/app/src/main/java/com/android/feedme/model/data/Profile.kt
+++ b/app/src/main/java/com/android/feedme/model/data/Profile.kt
@@ -12,5 +12,6 @@ data class Profile(
     val filter: List<String> = listOf(), // Setting of allergies / setting
     val recipeList: List<String> = listOf(), // Assuming this is a list of recipe IDs
     var savedRecipes: List<String> = listOf(), // Assuming this is a list of recipe IDs
-    val commentList: List<String> = listOf()
+    val commentList: List<String> = listOf(),
+    var showDialog: Boolean = true,
 )
--- a/app/src/main/java/com/android/feedme/model/data/Profile.kt
+++ b/app/src/main/java/com/android/feedme/model/data/Profile.kt
@@ -13,5 +13,5 @@ data class Profile(
     val recipeList: List<String> = listOf(), // Assuming this is a list of recipe IDs
     var savedRecipes: List<String> = listOf(), // Assuming this is a list of recipe IDs
     val commentList: List<String> = listOf(),
-    var showDialog: Boolean = true,
+    var showDialog: Boolean = true, // Shows or hides the pop-up in the generation of recipes
 )
--- a/app/src/main/java/com/android/feedme/ui/find/FindRecipeScreen.kt
+++ b/app/src/main/java/com/android/feedme/ui/find/FindRecipeScreen.kt
@@ -24,6 +24,7 @@ import androidx.compose.material3.Switch
 import androidx.compose.material3.SwitchDefaults
 import androidx.compose.material3.Text
 import androidx.compose.runtime.Composable
+import androidx.compose.runtime.collectAsState
 import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.remember
 import androidx.compose.ui.Alignment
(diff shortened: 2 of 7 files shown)


xrsula:
I fixed the popup system:
When the user opens the generate recipe page for the first time, the popup apparates to explain the system of Strict vs. Extra ingredients in the recipes that will be generated. The user can choose to either not show the popup again or not. 
There is a toggle underneath the "Ingredients" title that lets the user choose between a Strict vs Extra generation, and they can generate the recipes by pressing on the check button in the bottom right

This is the appearance of the popup and toggle:

https://github.com/feedme-swent-epfl/feedme-android/assets/100283160/28100166-8735-40df-a49f-34ad0d2590a4

app/src/main/java/com/android/feedme/model/data/Profile.kt
```diff
@@ -12,5 +12,6 @@ data class Profile(
     val filter: List<String> = listOf(), // Setting of allergies / setting
     val recipeList: List<String> = listOf(), // Assuming this is a list of recipe IDs
     var savedRecipes: List<String> = listOf(), // Assuming this is a list of recipe IDs
-    val commentList: List<String> = listOf()
+    val commentList: List<String> = listOf(),
+    var showDialog: Boolean = true,
```
JeannedeM, line 16: Maybe change `showDialog` to a more descriptive name like `showGeneratingModeInfos` or something else like that. It will make future changes or debugging easier :)

app/src/main/java/com/android/feedme/ui/find/FindRecipeScreen.kt
```diff
@@ -1,42 +1,54 @@
...
 import com.android.feedme.ui.navigation.Route
 import com.android.feedme.ui.navigation.Screen
 import com.android.feedme.ui.navigation.TOP_LEVEL_DESTINATIONS
 import com.android.feedme.ui.navigation.TopBarNavigation
+import com.android.feedme.ui.theme.DarkGrey
 import com.android.feedme.ui.theme.FabColor
+import com.android.feedme.ui.theme.OffWhite
 import com.android.feedme.ui.theme.TextBarColor
 
 /**
```
JeannedeM, line 54: Can you just add a very small sentence in the documention to make it more complete & descriptive :)

app/src/main/java/com/android/feedme/ui/find/FindRecipeScreen.kt
```diff
@@ -98,17 +171,48 @@ fun FindRecipeScreen(navigationActions: NavigationActions, inputViewModel: Input
               )
 
               // Line separator
```
JeannedeM, line 173: Is this commented code gonna be removed ?

JeannedeM (commented):
Geat job ! Thank you for all the modifications you made to reach the final product <3

app/src/main/java/com/android/feedme/model/data/Profile.kt
xrsula: Okay will change that ! :)

app/src/main/java/com/android/feedme/ui/find/FindRecipeScreen.kt
xrsula: Yes ! I removed that

app/src/main/java/com/android/feedme/ui/find/FindRecipeScreen.kt
JeannedeM: ok i think it might be a display problem, don't worry about it :) it's just that in the documentation of the function I only see the @param but no explenation of what the function is doing. But it might be a problem coming from me....

xrsula: Oh okay 😭  all good then :)

JeannedeM (approved):

PR #235, opened 2024-05-23
Fix : Recipe missing data, not well formated ingredient

fdejean:
Find a fix such that our scrapped recipes do not lack 

- [x] description
- [x] rating
- [x] Ingredient Quantity
- [x] Ingredient Dose


fdejean:
I filter the best 200 recipes, then I asked Chat GPT to fill the gap on the missing data.
The notebook also have method to set a recipe, update the userId, filter them by titles, get all the recipes for the database

fdejean:
Note for reviewer, you can see the file correctly formatted  if you click on "view file".

xrsula (approved):
Everything looks clean ! The scripts are pretty ingenious, exploiting the skills of ChatGPT really helps make our recipes more robust. The modifications in the view models also make sense, it is better to have global variables , looks cleaner :)
Well done !

PR #236, opened 2024-05-23
Fix add image url to database recipe

fdejean:
Issue : our scrapped recipe does not have an image url...
This PR aim at finding a solution

### DIFF
--- a/scraping/upload_img.py
+++ b/scraping/upload_img.py
@@ -1,0 +1,124 @@
+import json
+import requests
+from PIL import Image
+from io import BytesIO
+import firebase_admin
+from firebase_admin import credentials, firestore, storage
+
+# Initialize Firebase Firestore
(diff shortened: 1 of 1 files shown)


fdejean:
Tried to make a script that will search for the first image matching the title, 
I could not easily get the image url...

fdejean:
As the recipes in the database is only 200, the team will manually add the image url.
Hence I made a script that prompt user to enter a valid image url that match the given recipe title
the script will

- [ ]  check if the image url exist and download it
- [ ] Resize it to 16/9 aspect ratio for our thumbail
- [ ] Upload it to our firebase cloud
- [ ] Update the recipe url attribute with the url to the uploaded image
- [ ] Prioritize highest rated recipe as they are more relevant

JeannedeM (approved):
Thank you for creating this script that's really amazing to solve our images problem !!!

PR #237, opened 2024-05-23
Navigation bugs fix

ohamz:
These bugs were caused by fetching the profile of the author of a recipe.

- [x] Fix bug where app crashes when clicking on RecipeCard from HOME page (of recipe other than lasagna1 or pasta1).
- [x] Fix bug where app crashes when clicking on RecipeCard from SAVED page (of recipe other than lasagna1 or pasta1).
- [x] Fix bug where Omar's profile overrides user profile after clicking on lasagna1 or pasta1 then going to PROFILE page.


ohamz:
Ready for review :)

app/src/androidTest/java/com/android/feedme/test/LandingTest.kt
mche3kek, line 1: Nice to see fixed things, everything is coming together I am a proud dad 😍

mche3kek (commented):
Everything looks good to me code-wise ! The bugs are indeed fixed as specified :)

I just noticed some minor bug fixes in the UI and I think it would be great to address them (maybe not in this PR, let me know if you think we should have a separate one) : 

 - the rating in the FullRecipeDisplay is not centered, which kind of bothers me
image

 - as we have a share button on the thumbnail, it would be great to add it alongside the rating in a row and center it properly, what do you think of that ?

app/src/androidTest/java/com/android/feedme/test/LandingTest.kt
ohamz: hehee :)

ohamz:
Nice catch ! Yes I agree that these small UI features should be fixed, we can raise it as an issue for Sprint 10.

mche3kek (approved):

PR #238, opened 2024-05-23
Arranging Ingredients Input Screens

mche3kek:
Adding an intermediate screen used to display the picture taken / chosen from the gallery. This screen contains the Barcode and TextRecognition buttons and returns to the camera screen once the analyzing is done displaying either an error message or a success message. This modification is for a more user-friendly and understandable experience instead of the previous one where the user doesn't understand that he needs to take a picture before pressing on the analyzing button.


app/src/main/java/com/android/feedme/model/viewmodel/CameraViewModel.kt
```diff
@@ -324,4 +315,18 @@ class CameraViewModel : ViewModel() {
...
+   */
+  fun empty() {
+    _errorToDisplay.value = null
+    _informationToDisplay.value = null
+  }
+
+  /** This function is used to empty the list of ingredients after it has been processed. */
+  fun emptyIngredients() {
+    _listOfIngredientToInput.value = emptyList()
+  }
```
ohamz, line 331: Nice idea to add inner boolean values, to be sure nothing is executed before the screen is displayed.

app/src/main/java/com/android/feedme/ui/camera/Camera.kt
```diff
@@ -82,38 +67,27 @@ import com.android.feedme.ui.theme.CameraButtonsBackground
  * images and viewing them in a gallery. Utilizes CameraX for camera operations and Jetpack Compose
  * for the UI components.
  */
-@OptIn(ExperimentalMaterial3Api::class)
+@OptIn(ExperimentalMaterial3Api::class, ExperimentalCoroutinesApi::class)
```
ohamz, line 70: What does this add-on do ?

app/src/main/java/com/android/feedme/ui/camera/Camera.kt
```diff
@@ -321,28 +280,3 @@ fun CameraPreview(controller: LifecycleCameraController, modifier: Modifier = Mo
...
-        modifier = modifier) {
-          items(bitmaps) { bitmap ->
-            Image(
-                bitmap = bitmap.asImageBitmap(),
-                contentDescription = "Photo",
-                modifier = Modifier.clip(RoundedCornerShape(10.dp)))
-          }
-        }
-  }
-}
```
ohamz, line 348: So the local gallery feature was totally removed right ?

ohamz (changes requested):
Great coverage ! Implementing and testing this issue was not simple, well done!

I like the new division of the buttons through 2 screens, and the navigation is smooth.

I only have a few questions below.

app/src/main/java/com/android/feedme/ui/camera/Camera.kt
mche3kek: Yes indeed, it was not needed now that we process pictures one by one

app/src/main/java/com/android/feedme/ui/camera/Camera.kt
mche3kek: It was a remnant of something I tried using coroutines to wait for the result of the photo before navigating to the second screen, before settling with using Booleans. I will remove it :)

ohamz (approved):
Great. Approved !

PR #251, opened 2024-05-26
230 google ml kit  image labelling

JeannedeM:
Requirements :

- [x]     A button for image scanning is present in analyze photo screen.
- [x]     User can extract an ingredient name from a picture.
- [x]     The extracted ingredient is added to ingredient list.
- [x]     Success message is displayed in a green snackbar.
- [x]     Relevant error messages are displayed in a red snackbar.

Notes : 
The possible TensorFlow model usable were :

-    [mobile_object_labeler_v1]( https://www.kaggle.com/models/google/mobile-object-labeler-v1)
-    [aiy]( https://www.kaggle.com/models/google/aiy)

After some research and testing : 

- **First model :** this the one implemented. It's not ideal because it's supposed to be used to classify object in general sense (nike, curtains, etc...). However we are gonna use it while only considering labels of edible item. It gives approximetly 25-30 usable lables.
- **Second model** : turnsout this model is for already cooked meals classification (lasagna, cake, etc) => absolutly not what we want.

Tests : 
(Even though the model is able to detect and label up to 5 objects in photo, please take a photo of only one object)

- a white egg -> worked well
- a banana -> solala
- a green apple -> only detecting food or fruit 😢 

Of course this restricted amount of labels and this not really adpated model is not ideal. But with the short amount of time left it's not gonna be possible to do something better. It's fine for a beta version of the app. Training a model is a project in itself and make it compatible with the google ML kit is even more complicate.


app/src/main/java/com/android/feedme/model/viewmodel/CameraViewModel.kt
```diff
@@ -185,6 +214,59 @@ class CameraViewModel : ViewModel() {
...
+            },
+            onFailure = {
+              _errorToDisplay.value = ERROR_NO_OBJECT
+              errorObjectOccurred = true
+              continuation.resume(null)
+            })
+      }
+    }
+        ?: run {
+          _errorToDisplay.value = ERR_TIMEOUT
```
FacerOfGod, line 265: I think this is the line of the error bug. You can start by looking here

FacerOfGod (changes requested):
Nice, this is amazing it recognizes some fruits !! Everything in the description has been added successfully. However, just while testing I noticed that some of the error messages aren't correct. There might be some modifications to make regarding that. 

Bug case : When testing, taking a picture of a non food always sends a ERR_TIMEOUT and not a ERR_NO_FOOD_LABEL.

app/src/main/java/com/android/feedme/model/viewmodel/CameraViewModel.kt
JeannedeM: Ok I'm fixing that now ! Thank you :)

FacerOfGod (approved):
All good for me.

Even if the coverage isn't enough, the team has decided that the time and effort to build these tests are not going to be worth it so you can merge :D

PR #253, opened 2024-05-26
245 UI fix  fix the rating part in full recipe display  description of the recipe

JeannedeM:
Tasks :

- [x]    Add description of recipe to full display.
- [x]     Fix rating position on recipe full display.
- [x]     Fix user name position on full display.


app/src/main/java/com/android/feedme/ui/component/RecipeFullDisplay.kt
```diff
@@ -201,17 +201,18 @@ fun GeneralInfoDisplay(
...
+                            profileViewModel.setViewingProfile(profile)
+                            navigationActions.navigateTo(Screen.PROFILE)
+                          }),
+              text = "@${profile.username}",
               textAlign = TextAlign.Center,
               color = BlueUsername,
               style = TextStyle(fontSize = 16.sp, fontWeight = FontWeight.Medium))
         }
-        Spacer(modifier = Modifier.weight(1f))
+        // Spacer(modifier = Modifier.weight(1f))
```
ohamz, line 215: I find the username and the rating really close, why did you remove this spacer ?

image

ohamz: Yes I think it's better with the Spacer.


> people found a bit weird the way user name and rating was display.

Actually I think people found it weird when there is only the ratins displayed (when no author for the recipe is found), so maybe add the Spacer to the if statement of when the profile username is displayed ?

JeannedeM: If there is no user name, the rating will be centered like this (i moved the spacer inside the if, like you said).
image

0marwa: I think that's a good solution!

0marwa (approved):
This fix seems to have cleaned up the whole username/rating display situation, good job! 
And the UI still matches the original Figma design with the description added. The code is clean, so good job (:

PR #254, opened 2024-05-28
Expanding the GalleryLauncher to be able to handle all needed tasks correctly

mche3kek:
Implementing a pattern matching in the photo picker to handle all the use cases for the gallery launcher.
The comment case will be handled by @0marwa in another PR.

Fixing a bug in the profile picture display of other users and linking all occurences of profile pictures with the correct profile URL


JeannedeM (approved):
Great Work, thank you for this new feature ! All good for me :)

PR #262, opened 2024-05-31
Search screen fixes

ohamz:
This PR will solve the following fixes
- [x] Clicking on a RecipeCard in the Search Screen doesn't display it
- [x] Load more button on search screen wasn't loading more Recipes
- [x] Load more button is displayed even when there are no more recipes/profiles left to be displayed => Fix implemented: the Button is not displayed after clicking at it once.

### DIFF
--- a/app/src/androidTest/java/com/android/feedme/screen/SearchScreen.kt
+++ b/app/src/androidTest/java/com/android/feedme/screen/SearchScreen.kt
@@ -18,4 +18,5 @@ class SearchScreen(semanticsProvider: SemanticsNodeInteractionsProvider) :
   val noRecipesText: KNode = child { hasTestTag("NoRecipes") }
   val noAccountsText: KNode = child { hasTestTag("NoAccounts") }
   val recipeCard: KNode = child { hasTestTag("RecipeCard") }
+  val loadMoreButton: KNode = child { hasTestTag("Load More Button") }
 }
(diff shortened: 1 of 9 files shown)


FacerOfGod (changes requested):
Good changes!! Here is a summary for each bug : 
- **RecipeCard:** I tried a lot of Recipes, using different searches and accounts. I couldn't reproduce this bug. I consider It is solved.
- **Load button not working:** Same with this bug. I couldn't reproduce this bug. I consider It is solved
- **Load button displayed** Your fix was well implemented, tested it and it has expected behaviour as for the profiles. However I think we could solve it differently. Imagine, we fetchand limit the database to X recipe each time. Can we just check : 
_1 - If the result is under X -> don't show the button
2 - if the result is greater than X-1 -> show only X-1 recipe/profile and the button_
So we will always fetch one more recipe/profile than the ones we show, but on every click of the button at least one more element should pop out. What do you think ?

ohamz:
However I think we could solve it differently. Imagine, we fetchand limit the database to X recipe each time. Can we just check :
>   _1 - If the result is under X -> don't show the button
>   2 - if the result is greater than X-1 -> show only X-1 recipe/profile and the button_
>   So we will always fetch one more recipe/profile than the ones we show, but on every click of the button at least one more element should pop out. What do you think ?

I thought of implementing this change, however, I found something missing. What if we have exactly X recipes available that are fetched ? The button would still show, and with the way the fetching from the database is implemented, when clicking on it it would redisplay the recipes displayed at the start (and would continue each time we click it).

ohamz:
Retested functionalities after rebasing and it still works.

ohamz:
@FacerOfGod I agree that this implementation would fit better. However, I don't think that the implementation will be trivial.
I suggest for it to be done in another PR, but I don't know if I will have the time since I already spent more than 25 hours on the project this week.

FacerOfGod:
@ohamz Yes of course completely understandable. Perfect than all good for me !

FacerOfGod (approved):

PR #266, opened 2024-06-01
IngredientList bugs and reduce complexity update

FacerOfGod:
Cleaning up the IngredientList file to match the requirements in term of code complexity and fix UI bugs

 **Complexity :**

- [x] Remove InputIngredientState system way to confusing to use 
- [x] Reduce cognitive complexity

** Bug Fix :**

- [x] Quantity : if completely empty crashes the app
- [x] Quantity : if modifed last you can not validate the ingredient
- [x] Ingredient is considered as a valid ingredient in the database aka. green but should be red
- [x] Ingredient is case sensitive, not a great user experience
- [x] Generate button was clickable even if there was empty fields

### DIFF
--- a/app/src/main/java/com/android/feedme/model/viewmodel/InputViewModel.kt
+++ b/app/src/main/java/com/android/feedme/model/viewmodel/InputViewModel.kt
@@ -7,7 +7,6 @@ import androidx.security.crypto.EncryptedSharedPreferences
 import androidx.security.crypto.MasterKey
 import com.android.feedme.model.data.IngredientMetaData
 import com.android.feedme.model.data.MeasureUnit
-import com.android.feedme.ui.component.IngredientInputState
 import com.google.gson.Gson
 import com.google.gson.reflect.TypeToken
 import kotlinx.coroutines.flow.MutableStateFlow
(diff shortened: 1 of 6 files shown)


app/src/main/java/com/android/feedme/model/viewmodel/InputViewModel.kt
```diff
@@ -35,15 +34,18 @@ class InputViewModel(val context: Context? = null) : ViewModel() {
   val listOfIngredientMetadatas: StateFlow<List<IngredientMetaData?>> = _listOfIngredientMetadatas
 
   // StateFlow to hold the total number of complete ingredients
-  private val _totalCompleteIngredientMetadatas = MutableStateFlow(0)
-  val totalCompleteIngredientMetadatas: StateFlow<Int> = _totalCompleteIngredientMetadatas
+  private val _isComplete =
+      MutableStateFlow(
+          _listOfIngredientMetadatas.value.count { isCompleteIngredient(it) } ==
+              _totalIngredientEntriesDisplayed.value - 1)
```
xrsula, line 40: I don't fully understand what you mean by _totalIngredientEntriesDisplayed.value -1 ?

xrsula (approved):
Everything looks okay to me, the code is logical for the most part and is cleaned up nice. I was able to test the code to check if the bugs mentionned were fixed, which is the case.
Nice job :)

app/src/main/java/com/android/feedme/model/viewmodel/InputViewModel.kt
0marwa: I think the `-1` is there because our list is expected to have at least one `null` element, therefore we check if our count of complete ingredients in the list matches the number of ingredients to display minus the `null` element. correct me if i'm wrong @FacerOfGod !

0marwa (approved):
good job spotting and fixing the existing bugs, the user flow is much nicer / logical now, and the input handling is very secure now! i checked out the branch and the code functions as expected from the list of fixes you mentioned. it's also clean and well-documented as well, so all good for me too (:

app/src/main/java/com/android/feedme/model/viewmodel/InputViewModel.kt
FacerOfGod: That is correct :)

PR #268, opened 2024-06-02
Adding tags for recipe difficulty

0marwa:
Adding tags to the recipe to specify how hard it is to cook.


mche3kek (commented):
image
image

I'm adding screenshots for better documentation. Good job on this feature ! I really like the tag, it feels very nice. 

I think that it would look better if we center the difficulty level in the full display of the recipe, what do you think ?

xrsula:
> level in the full display of the recipe, wh

I think it is more coherent to keep it that way, it is more homogeneous with the rest of the text that is displayed :)

mche3kek (approved):

PR #270, opened 2024-06-02
Bug thumnbail display the same if the profile has no created recipe

fdejean:
Side fix : if a user had no follower, it did not display them correctly as there was a null check before the fetch.
Instead step the state flow to null directly


ohamz (approved):
Awesome ! Thank you for your quick response.

PR #273, opened 2024-06-02
Last minute bug should be fixed

### LINKED ISSUE #271: Last minute Input bug
**Goal :** Solve UI update bugs

**Bugs :**

- [x] When manually inputting text, app would randomly pick the users ingredient
- [x]  With camera every Ingredient was multiplied by two

These bugs don't appear in the emulator for some reason so was hard to find them

FacerOfGod:
**Goal :** Solve UI update bugs

**Bugs :**
- [x] When manually inputting text, app would randomly pick the users ingredient
- [x] With camera every Ingredient was multiplied by two

These bugs don't appear in the emulator for some reason so was hard to find them


mche3kek (approved):
Good to go !

JeannedeM (approved):
Thanks for fixing this bug so fast and at the last minutes ! The interface is way better and user friendly now :)

PR #275, opened 2024-06-02
clear all output of notebook

fdejean:
The output count as lines, hence we are removing them to make our repo a kotlin project back


mche3kek (approved):
All good for me, thanks for your work !

